PR #1, opened 2014-08-01
7.0 add account statement coda import

lmignon:


account_statement_coda_import/__openerp__.py
```diff
@@ -0,0 +1,57 @@
...
+This module brings generic methods and fields on bank statement to deal with
+the importation of coded statement of account from electronic files. **CODA**
+
+This module allows you to import your bank transactions with a standard **CODA**
+file (you'll find samples in the 'data' folder). It respects the chosen profile
+(model provided by the account_statement_ext module) to generate the entries.
+
+ .. important::
+   The module requires the python library
+   `pycoda <https://pypi.python.org/pypi/pycoda>`_
```
sbidoul, line 49: pycoda should be in external_dependencies

account_statement_coda_import/__openerp__.py
```diff
@@ -0,0 +1,57 @@
...
+     'account_statement_base_import',
+     'account_statement_bankaccount_completion'
+     ],
+ 'description': """
+Bank statement CODA import
+==========================
+
+This module brings generic methods and fields on bank statement to deal with
+the importation of coded statement of account from electronic files. **CODA**
+
```
sbidoul, line 38: Possibly explain the module is an alternative to the official l10n_be_coda that leverages the advanced bank statement completion framework of the bank-statement-reconcile OCA project (https://github.com/OCA/bank-statement-reconcile)

account_statement_coda_import/__openerp__.py
lmignon, line 49: but if I put it in the external_dependencies, it becomes required to start the server even if the module is not installed.. I don't know what is the best...

account_statement_coda_import/parser/coda_file_parser.py
```diff
@@ -0,0 +1,192 @@
...
+    def _validate(self, *args, **kwargs):
+        """
+        No validation needed for this parser
+        """
+        return True
+
+    def get_st_vals(self):
+        """
+        This method return a dict of vals that ca be passed to
+        create method of statement.
```
sbidoul, line 97: spelling: that can be passed

account_statement_coda_import/statement.py
```diff
@@ -0,0 +1,106 @@
...
+                    'CODA based transaction'))
+        statement_line_obj = self.pool.get('account.bank.statement.line')
+        if'coda_account_number' in statement_line_obj._columns:
+            # True if l10n_be_coda is installed
+            res.append((L10NBECodaFileParser.parser_name,
+                        'CODA based transaction (L10nBECoda installed)'))
+        return res
+
+    _columns = {
+        'import_type': fields.selection(
```
sbidoul: I think it's not necessary to redefine the field anymore since get_import_type_selection is called by _get_import_type_selection: https://github.com/OCA/bank-statement-reconcile/blob/8.0/__unported__/account_statement_base_import/statement.py#L50

sbidoul:
:+1: code LGTM, tested and working in production

qdp-odoo:
Hello there,

in master, a recent commit introduce a generic wizard for importing bank statements where l10n_be_coda has just to 
*add 'coda' as a available file format
*define a function process_coda that receives as parameter a file and returns a list of values for the bank statement creation.

It could be nice if the official module could reuse the pycoda library, in the same way as we did for ofxparse library in account_bank_statement_import_ofx, and reduce in the same way the divergence between the 2 versions (the official one and yours) of the module. Any tip on achieving this?

In particular, you would like to 
*reuse _detect_partner() with the bank account of the partner, and store the result on the bank statement line 'bank_account_id' and 'partner_id' fields
*store the structured communication in 'ref' field, while the normal communications go in 'name' field, for automatic detection of the invoice in the bank statement reconciliation interface.

Best regards,

sbidoul:
Hello Quentin,

A bank statement import framework in the core is very interesting. Thanks for the pointers.

We'll certainly work into making the different approaches converge. I understand it is for 9.0 not 8.0, right?

qdp-odoo:
it's not for v8, but as it's already in master it will be released in the next saas version after the 8.0.

laetitia-gangloff:
test ok :+1:

Cedric-Pigeon:
OK for me :+1:

PR #8, opened 2014-11-14
[REF] auto-fix conventions

moylop260:


account_statement_coda_import/parser/coda_file_parser.py
```diff
@@ -157,6 +158,7 @@ def get_st_line_vals(self, line, *args, **kwargs):
 
 
 class L10NBECodaFileParser(CodaFileParser):
+
```
sbidoul, line 161: @moylop260 why  inserting a blank line here?

lmignon, line 161: @moylop260 I think you use an outdated autopep8 script. I observed the same phenomenon with an older version of autopep8.

sbidoul:
superseded by #9

PR #9, opened 2014-11-15
saturday housekeeping

sbidoul:


lmignon:
:+1:  And thank you for cleaning!

sbidoul:
Fast-tracking as it is only cleaning.

PR #12, opened 2015-01-04
add l10n_be_coa_multilang

luc-demeyer:


l10n_account_translate/README.rst
```diff
@@ -0,0 +1,7 @@
+Description
+===========
+
+This module sets translate=True on the account.account,name field
+and is intended for use by accounting localisation modules.
```
sbidoul: Is this not the same purpose as l10n_multilang in official 8.0?

luc-demeyer: L10n_multilang changes a whole set fields into multilingual fields whereas the l10n_account_translate only changes the account.account,name field. 
We are now implementing accounting on OpenERP for 5 years and for most of the customers we turn off the account.account,name translate=True since it's annoying when adding new accounts via 'duplicate'.

luc-demeyer:
I have signed the CLA agreements (for myself and for my company) half a year ago and sent them to luc maurer.

gurneyalex:
@luc-demeyer I'm currently investigation what is wrong in the bot.

gurneyalex:
@luc-demeyer: the problem was that your github login was not filled in the OCA database. This is now fixed.

sbidoul:
@luc-demeyer since l10n_account_translate is not Belgium-specific can your remove it from this PR and submit it to OCA/account-financial-tools? Thanks!

luc-demeyer:
@sbidoul 
I prefer we proceed in steps: 
Step 1: get both modules merged into the Belgian branch (should go fast).
Step 2: When step 1 is completed, I'll make a PR for account-financial-tools (which may take a while before it gets merged). 
Step 3: When step 2 is completed, I'll make a PR to remove the l10n_account_translate from the Belgian branch.

sbidoul:
Hi Luc,

Regardfing l10n_account_translate, we have a duty to keep things as well ordered as possible in OCA repositories. Since it is not a dependency of l10n_be_coa_multilang, it's quite easy to remove it from this PR and resubmit it to OCA/account-financial-tools.

While talking about dependencies, could you remove the dependency on account_cancel? Several people prefer to not use it, especially in combination with account_default_draft_move. So it's better it remains optional.

Thanks!

luc-demeyer:
@sbidoul 
cf. https://github.com/OCA/account-financial-tools/pull/128 to add l10n_account_translate to account_financial-tools.
Remark: of course I agree that we need to keep the OCA repos ordered, the debate is about when we do what.In this case there is a small issue: the l10n_be_coa_multilang setup wizard will fail when the l10n_account_translate module is not present. Therefor I included it into this PR. Once the module is merged into the financial tools we can remove it from this PR (unless you know a way to make a single PR against two different repos).

l10n_be_coa_multilang/account_financial_report.py
```diff
@@ -0,0 +1,293 @@
...
+        create/update the entries in the BNB/NBB legal report scheme.
+        """
+        note = ''
+        upd_ctx = {'update_be_reportscheme': True}
+        scheme_table = self.env['be.legal.financial.reportscheme'].search([])
+        be_report_ids = [x.report_id.id for x in scheme_table]
+        accounts = self.env['account.account'].search(
+            ['|', ('type', '!=', 'view'), '&', ('type', '=', 'view'),
+             ('centralized', '=', True)]
+            ).with_context(upd_ctx)
```
Olivier-LAURENT: It seems this `with_context` is missplaced (i.e. the code in `account.search()` related to this specific context is never executed). Probably you want to say  `self.env['account.account'].with_context(upd_ctx).search(...)`

luc-demeyer:
@Olivier-LAURENT 
Olivier,
thanks for reporting the with_context bug. I fixed this and at the same time updated the description to add OCA & contributors.

l10n_be_coa_multilang/wizard/accounting_report_wizard.py
```diff
@@ -0,0 +1,167 @@
...
+    def set_context(self, objects, data, ids, report_type=None):
+        report_date = datetime_field.context_timestamp(
+            self.cr, self.uid,
+            datetime.now(), self.context
+            ).strftime(DEFAULT_SERVER_DATETIME_FORMAT)
+        self.localcontext.update({'report_date': report_date})
+        super(report_financial_parser, self).set_context(
+            objects, data, ids, report_type)
+
+    def get_lines(self, data):
```
lmignon, line 67: This method potentially breaks other modules extending the same method. You should call the base implementation and enrich the data  received.

luc-demeyer, line 67: What exactly is the potential problem ?

lmignon, line 67: since you never call super, you ignore the others overrides

sbidoul:
setting WIP pending refactoring

l10n_be_coa_multilang/wizard/l10n_be_vat_declaration.py
```diff
@@ -0,0 +1,311 @@
...
+            'comments': comments,
+        }
+
+        if tax_info.get('VI')[1] >= 0:
+            tax_info['71'] = tax_info['VI']
+        else:
+            tax_info['72'] = tax_info['VI']
+        cases_list = []
+        for item in tax_info:
+            if tax_info['91'][1] and ending_month != 12:
```
adrienpeiffer, line 188: `ending_month != '12'` ?
cc\ @sbidoul

luc-demeyer:
This check could be refined, should be 12 and date <= december 20th.
I'll update the PR.
Cf documentation (in dutch) on vat case 91:

Betreft uitsluitend de maandaangifte van december
Werkelijk verschuldigde btw voor de periode van 1 tot 20 december

luc-demeyer:
Please forget my previous comment:

> > This check could be refined, should be 12 and date <= december 20th.

I reacted too fast (without looking into the code).
The current code should remain unchanged.
An amount on code 91 is only valid for the december declaration (which is ending_month 12).

pedrobaeza:
As discussed in the sprint, this is going to be worked on v9, so I close.

PR #18, opened 2015-03-08
8.0 Add module l10n_be_account_bank_statement_import_bpost

alexis-via:


pedrobaeza:
You should add new repo dependency on Travis.

The rest seems :+1: technically. Nothing to say about format itself.

l10n_be_account_bank_statement_import_bpost/account_bank_statement_import_bpost.py
```diff
@@ -0,0 +1,97 @@
...
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+import logging
+from datetime import datetime
+from StringIO import StringIO
```
lmignon: from cStringIO

l10n_be_account_bank_statement_import_bpost/account_bank_statement_import_bpost.py
```diff
@@ -0,0 +1,97 @@
...
+                cr, uid, data_file, context=context)
+        transactions = []
+        i = 0
+        account_number = currency_code = False
+        start_date_str = end_date_str = False
+        diff = 0.0
+        for line in unicodecsv.reader(
+                StringIO(data_file), encoding='utf-8', delimiter=';'):
+            i += 1
+            _logger.debug("Line %d: %s" % (i, line))
```
lmignon: _logger.debug(Line %d: %s" , i, line)

l10n_be_account_bank_statement_import_bpost/account_bank_statement_import_bpost.py
```diff
@@ -0,0 +1,97 @@
...
+
+_logger = logging.getLogger(__name__)
+
+
+class AccountBankStatementImport(models.TransientModel):
+    _inherit = 'account.bank.statement.import'
+
+    def _check_bpost(self, cr, uid, data_file, context=None):
+        return data_file[3:].startswith(
+            u'"Numéro de compte :";"'.encode('utf-8'))
```
lmignon: What happens if the file is in dutch?

l10n_be_account_bank_statement_import_bpost/account_bank_statement_import_bpost.py
```diff
@@ -0,0 +1,97 @@
...
+            if line[8]:
+                name += ' - ' + line[8]
+            if line[9]:
+                name += ' ' + line[9]
+            date = datetime.strptime(line[1], '%d-%m-%Y')
+            date_str = fields.Date.to_string(date)
+            amount = float(line[3].replace(',', '.'))
+            diff += amount
+            currency_code = line[4]
+            if not start_date_str and not end_date_str:
```
lmignon: Would it not be better to compare dates and not strings?

lmignon:
@alexis-via Thank you for the contribution. Is the Bport format a normalized format described in a publicly available document?  As a new module, it would be nice to have a unittest. 

Regards,

lmi

l10n_be_account_bank_statement_import_bpost/account_bank_statement_import_bpost.py
alexis-via: This code will need to be adapted if the file is in Dutch. But I don't have an example of a file in Dutch, because my customer is French.

l10n_be_account_bank_statement_import_bpost/account_bank_statement_import_bpost.py
alexis-via: Why use a C implementation when the Python implementation works very well ? :)

lmignon: Why not use a faster version of StringIO? https://docs.python.org/2/library/stringio.html#module-cStringIO

lmignon:
@alexis-via You need also to update .travis.yml..

``` yaml
...

install:
  - git clone https://github.com/OCA/maintainer-quality-tools.git ${HOME}/maintainer-quality-tools
  - git clone https://github.com/OCA/account-financial-reporting ${HOME}/account-financial-reporting -b ${VERSION}
  - git clone https://github.com/OCA/bank-statement-import ${HOME}/bank-statement-import -b ${VERSION
  - export PATH=${HOME}/maintainer-quality-tools/travis:${PATH}
  - travis_install_nightly
  - pip install xlrd pycoda

...
```

sbidoul:
Travis fails because of missing unicodecsv dependency.

@alexis-via can you add it to the pip install line in travis?

sbidoul:
@alexis-via if you don't have time to write a test, it would be great if you could put a sample file in the test directory, so people can easily see what these files looks like.

alexis-via:
@sbidoul OK. I would need to carefully anonymize the file... I'll try to find time to do that.

luc-demeyer:
@alexis-via 
This seems a strange contribution to me since you can get your bpost electronic statements in CODA format which is a lot richer than csv. 
What's the reason not to go for the (currently) most used format in Belgium ?

alexis-via:
@luc-demeyer 
Access to the CODA format is an option that you need to pay for ; access to the CSV format is free.

luc-demeyer:
For most of the major banks coda access is 'free' (included their business packs), 
When it's not free, than the price is usually a lot cheaper than the development & maintenance of a new import format, hence it remains a strange 'business requirement' to me since I don't think the business case makes a financial sense.

gurneyalex:
needs updating to use  oca_dependencies.txt

alexis-via:
Rebased and updated to use oca_dependencies.txt

Do you think it will be merged one day, or do you prefer that I maintain this module outside of OCA (in that case, i'll create a dedicated project on akretion's github) ?

sbidoul:
@alexis-via I certainly have no objection to merge this. I'll check your last changes ASAP.

l10n_be_account_bank_statement_import_bpost/README.rst
```diff
@@ -0,0 +1,34 @@
+Import Bpost CSV bank statements
```
sbidoul, line 6: @alexis-via can you just update the README.rst structure according to last template?
https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst
Licence at the top, link to issue tracker and runbot.

I know it's painful to maintain those readmes. At some point they should become partly autogenerated.

sbidoul:
Could you just care for the README.rst and while you are at it add the CSV header in the README so it's easy to know which columns are expected. Then we can merge.

luc-demeyer:
My 2 cents:
I am active in the Belgian banking sector for 30 years (my first job was at a bank in 1985). Already from that time all major banks had CODA files and in the meanwhile I don't know of any bank in Belgium not supplying CODA files at a reasonable cost (in any case cheaper than maintaining a primitive csv import and the lack of automation compared to the much more advanced CODA processing). 
I therefor don't think that the Belgian OCA repo is the right place to promote this module.
First of all, the OCA members need to look after quality and quality to me means functional as well as technical. I don't think the functional quality is respected here since supporting csv bank statements in Belgium is like going back to the stone age.
Secondly I don't think that those Belgian partners who understand accounting (which tend to be the ones that contribute to the Belgian localisation) will ever use or recommend this module hence the Belgian OCA publication will only add burden to maintain this code but this burden will not be compensated by quality contributions.

pedrobaeza:
This is still red. Do you plan to continue on this?

alexis-via:
Travis is now green. Note that Travis is red on the main OCA/l10n-belgium/8.0 branch because mis-builder was missing in oca_dependencies.txt ; I had to fix this in this PR to make travis green.

sbidoul:
@alexis-via thanks for the update. Could you add the list of column headers in the README, so we know what this module expects? Can you also add a word saying that this module should be used only when it is not possible to obtain CODA files.

PR #19, opened 2015-03-16
[FIX] Tests with pycoda-0.4

lmignon:
In the latest version (0.4) of the coda parser (pycoda) comments from the details line are no more summurized in the globailsation


sbidoul:
:+1:

laetitia-gangloff:
:+1:

PR #20, opened 2015-03-17
Add module account_bank_statement_import_coda

lmignon:
This module is a rewrite of l10n_be_coda from odoo master using the pycoda parser.
Expected benefits:
- The pycoda parser is a better-tested external python lib.
- The module only depends of `account_bank_statement_import` (`l10_be_coda` on master branch depends of `account_voucher`, `base_iban`, `l10n_be_invoice_bba`, `account_bank_statement_import`)
- Better separation between file parsing and mapping of these data in Odoo.
- [x] This PR depends of https://github.com/OCA/bank-statement-import/pull/11


lmignon:
The tests fail due to https://github.com/OCA/maintainer-quality-tools/pull/169

account_bank_statement_import_coda/__openerp__.py
```diff
@@ -0,0 +1,41 @@
...
+{
+    'name': 'Import CODA Bank Statement',
+    'author': "ACSONE SA/NV,Odoo Community Association (OCA)",
+    'website': "http://www.acsone.eu",
+    'category': 'Accounting & Finance',
+    'version': '1.0',
+    'license': 'AGPL-3',
+    'depends': [
+        'account_bank_statement_import',
+        'l10n_be_coda'
```
sbidoul: Do we really need this dependency on l10n_be_coda?

account_bank_statement_import_coda/wizard/account_bank_statement_import_coda.py
```diff
@@ -0,0 +1,218 @@
...
+                'transactions': transactions
+                }
+        name = statement.paper_seq_number
+        if name:
+            year = ""
+            if statement.creation_date:
+                parsed_date = date_parser.parse(statement.creation_date)
+                year = "%s/" % parsed_date.year
+            vals.update({
+                'name': "%s%s" % (year, statement.paper_seq_number),
```
sbidoul, line 98: Add the journal code as prefix?

account_bank_statement_import_coda/__openerp__.py
lmignon: If you need coda, you probably depend from l10n_be for your business. But if you install l10n_be, l10n_be_coda will be installed as a dependency of l10n_be. By depending of l10n_be_coda, we can safely share the same model, avoid to redefine the same field in the account_bank_statement view and hide the menu item that leads to 'old style' import wizard provided by l10n_be_coda.

account_bank_statement_import_coda/wizard/account_bank_statement_import_coda.py
lmignon, line 98: The journal info is not available at this stage of the process. Here we are on the first step of the import process.
If we need to add the journal code as prefix, we should override '_complete_stmts_vals'. If you think we need it, I can do it.

account_bank_statement_import_coda/__openerp__.py
sbidoul: Some don't use l10n_be and l10n_be_coa_multilang instead.

By same model, you mean the "CODA Notes" field? I always thought it was weird to have a specific field for that. I'd say it could be a generic field on statement or statement line that holds additional information and could be used by any parser. So not CODA specific.

lmignon: OK, I'll changes my code to remove the dependency on l10n_be_coda.
You're right for the "CODA Notes" field. I can maybe propose this new field upstream in account_banck_statement_import and use this new one.

account_bank_statement_import_coda/wizard/account_bank_statement_import_coda.py
```diff
@@ -0,0 +1,218 @@
...
+                filter(lambda l: l.type != MovementRecordType.GLOBALISATION,
+                       statement.movements)
+                ):
+            if line.type != MovementRecordType.GLOBALISATION:
+                info = self.get_st_line_vals(line,
+                                             globalisation_dict)
+                info['sequence'] = sequence
+                transactions.append(info)
+        coda_notes = []
+        for line in statement.informations:
```
sbidoul: In this module, 'line' has many different meanings at different places. It makes the code a bit difficult to understand.

account_bank_statement_import_coda/wizard/account_bank_statement_import_coda.py
```diff
@@ -0,0 +1,218 @@
...
+        if line.transaction_amount_sign == AmountSign.DEBIT:
+            amount = - amount
+        return {'name': self.get_st_line_name(line, globalisation_dict),
+                'date': line.entry_date or datetime.datetime.now().date(),
+                'amount': amount,
+                'ref': line.ref,
+                'partner_name': line.counterparty_name or None,
+                'account_number': line.counterparty_number or None,
+                'note': self.get_st_line_note_msg(line),
+                'unique_import_id': line.ref + line.transaction_ref,
```
sbidoul: Is line.ref a good element to make a unique line id? Perhaps the year + paper_seq_number + transaction_ref makes a more unique identifier?

account_bank_statement_import_coda/__openerp__.py
sbidoul: Having the note fields in account_bank_statement_import make sense to me. If upstream accepts.

sbidoul: I see there is already a note field on bank.statement.line in the account module. Do we need more?

account_bank_statement_import_coda/wizard/account_bank_statement_import_coda.py
```diff
@@ -0,0 +1,218 @@
...
+        balance_start = statement.old_balance
+        if statement.old_balance_amount_sign == AmountSign.DEBIT:
+            balance_start = - balance_start
+        balance_end_real = statement.new_balance
+        if statement.new_balance_amount_sign == AmountSign.DEBIT:
+            balance_end_real = - balance_end_real
+        transactions = []
+        vals = {'balance_start': balance_start,
+                'balance_end_real': balance_end_real,
+                'date': statement.creation_date,
```
sbidoul: Is new_balance_date a better date?

account_bank_statement_import_coda/wizard/account_bank_statement_import_coda.py
```diff
@@ -0,0 +1,218 @@
...
+class account_bank_statement_import(models.TransientModel):
+    _inherit = 'account.bank.statement.import'
+
+    def _check_coda(self, data_file):
+        if Parser is None:
+            return False
+        try:
+            # Matches the first 24 characters of a CODA file, as defined by
+            # the febelfin specifications
+            return re.match(r'0{5}\d{9}05[ D] {7}', data_file) is not None
```
sbidoul, line 36: The detection feature could be part of pycoda

account_bank_statement_import_coda/__openerp__.py
lmignon: @sbidoul according to the febelfin specification file, an information record refers to a movement record (IOW, the 'transaction_ref' of an information statement refers the 'transaction_ref' of a movement statement) . Therefore we have a way to store the communication message from information record in the field 'note' of the related bank.statement.line. I propose to ignore the information provided by the free communication records. These don't refer to movements.

sbidoul:
:+1:

lmignon:
rebased

sbidoul:
@lmignon the bank account used in the demo data is the same as the one in l10n_be_coda. Therefore the sanitized_acc_number constraint prevents to create it.

lmignon:
@sbidoul green after rebased

lmignon:
@sbidoul The problem occurred when configuring the accounting data to use the Belgium Chart of account. In this case, the l10n_be_coda is automatically installed and defined the same account number as the one in account_bank_statement_import_coda. To fix the problem I've changed the one defined in  account_bank_statement_import_coda.

sbidoul:
The IBAN checksum is not correct but I guess it's not an issue for this purpose.

:+1:

lmignon:
@sbidoul now it's valid
image

sbidoul:
@lmignon any idea why tests are failing here?

lmignon:
@sbidoul No idea. The search on the expected statement name return no result.

sbidoul:
I fixed the test failure (it was a real bug), and brought the module up to the latest OCA standard.

I think it's now ready to merge at last.

cc/ @lmignon @adrienpeiffer

sbidoul:
The test failure is unrelated. It's in the company_web module and apparently resolved by https://github.com/OCA/account-financial-reporting/pull/205.

adrienpeiffer:
:+1:

lmignon:
:+1:

PR #24, opened 2015-08-28
8.0 add transifex status lmi + missing badges

lmignon:


sbidoul:
:+1: I'm fast tracking as it is a maintenance operation.

PR #26, opened 2015-08-28
OCA Transbot updated translations from Transifex

lmignon:


yvaucher:
:+1:

pedrobaeza:
So translations are proposed as PR in your tx-pull branch or this is only a test?

lmignon:
@pedrobaeza The translations have been pushed by tx-pull to a test repository and I've created a PR from this repository. I've updated tx-pull to be able to provide an alternative org where to push the translations so it's possible for a user without commit right on a repository to get the translations and do a  manual PR with the translations...

pedrobaeza:
Ah, OK, thanks for the clarification.

sbidoul:
:+1:

adrienpeiffer:
:+1:

PR #30, opened 2016-01-15
Allow companyweb to be updated from external scripts

jrial:
This PR allows updating of companyweb data from scripts, e.g. using the oerplib library to connect to an Odoo instance. The PR is written against 9.0, but applies cleanly against the 7.0 and 8.0 branches as well.

When calling Odoo methods over xmlrpc, one can't pass None nor Browse Records; this change rectifies the API so that this never happens.

### DIFF
--- a/account_companyweb/model/res_partner.py
+++ b/account_companyweb/model/res_partner.py
@@ -37,6 +37,8 @@ class res_partner(orm.Model):
     _inherit = 'res.partner'
 
     def companyweb_information(self, cr, uid, ids, vat_number, context=None):
+        if not context:
+            context = {}
         login = self.pool['ir.config_parameter'].get_param(
             cr, uid, 'companyweb.login', False)
         pswd = self.pool['ir.config_parameter'].get_param(
--- a/account_companyweb/wizard/account_companyweb_wizard.py
+++ b/account_companyweb/wizard/account_companyweb_wizard.py
@@ -47,9 +47,10 @@ class account_companyweb_wizard(orm.TransientModel):
         'result': fields.float('Fiscal Year Profit/Loss (+/-)', readonly=True),
     }
 
-    def get_update_values(self, cr, uid, ids, wizard, context=None):
+    def get_update_values(self, cr, uid, ids, context=None):
         """ This method is designed to be inherited to add some field to
             update on res.partner"""
+        wizard = self.browse(cr, uid, ids)
         return {'name': wizard.name,
                 'is_company': True,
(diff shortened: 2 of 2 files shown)


account_companyweb/model/res_partner.py
```diff
@@ -37,6 +37,8 @@ class res_partner(orm.Model):
     _inherit = 'res.partner'
 
     def companyweb_information(self, cr, uid, ids, vat_number, context=None):
+        if not context:
+            context = {}
```
jrial, line 41: context is being returned at the end of the method. When calling it over xmlrpc, it can't be None as None can't be marshalled.

account_companyweb/wizard/account_companyweb_wizard.py
```diff
@@ -47,9 +47,10 @@ class account_companyweb_wizard(orm.TransientModel):
         'result': fields.float('Fiscal Year Profit/Loss (+/-)', readonly=True),
     }
 
-    def get_update_values(self, cr, uid, ids, wizard, context=None):
+    def get_update_values(self, cr, uid, ids, context=None):
```
jrial, line 50: Browse records can not be passed over xmlrpc when calling from scripts; the parameter ids provides sufficient information to retrieve the concerned wizard object.

account_companyweb/wizard/account_companyweb_wizard.py
```diff
@@ -62,7 +63,7 @@ def update_information(self, cr, uid, ids, context=None):
         res_partner_model = self.pool['res.partner']
         partner_id = context['active_id']
         this = self.browse(cr, uid, ids, context=context)[0]
-        update_values = self.get_update_values(cr, uid, ids, this,
+        update_values = self.get_update_values(cr, uid, ids,
```
jrial, line 66: And of course, update the method call to the new signature.

jrial:
Some context: I'm implementing this for a customer so that they can run a weekly cron job to update their companyweb data. One of the features I'm implementing is rate limiting (they have X amount of free requests per week), and for this I would add a datetime field to res.partner to indicate when it has last been updated. This field would be written every time the companyweb data is fetched for that particular partner.

Is this something of interest to the broader community? And while we're at it, would the cron job be of interest as well? If so, I could create pull requests for that as well.

jrial:
Closing this PR for now; I noticed some other related issues in the codebase that need to be fixed as well. Will create a new PR as soon as I get around to porting them.

PR #35, opened 2016-09-13
8.0 Add module l10n_be_invoice_ubl to support the e-fff standard

alexis-via:
Left todo:
- [ ] add more mandatory fields, cf http://www.e-fff.be/files/efff_UBL_Belgique_V110_Final.xls
- [ ] Check if we have to switch to UBL 2.0 instead of 2.1
- [ ] validate the files generated by Odoo ?

### DIFF
--- a/l10n_be_invoice_ubl/models/account_invoice.py
+++ b/l10n_be_invoice_ubl/models/account_invoice.py
@@ -39,7 +39,7 @@ class AccountInvoice(models.Model):
         '''Read from http://www.e-fff.be/FR/doc03.php :
         The e-fff community recommends to use the following file naming:
         efff_BE0123456789_AlphaNumericCharactersFreeOfChoice.xml'''
-        filename = 'efff_%s_Invoice_%s' % (
+        filename = 'efff_%s_Invoice_%s.xml' % (
             self.commercial_partner_id.sanitized_vat,
             self.number and self.number.replace('/', '-') or 'draft')
         return filename
--- a/l10n_be_invoice_ubl/models/account_invoice.py
+++ b/l10n_be_invoice_ubl/models/account_invoice.py
@@ -10,12 +10,16 @@ class AccountInvoice(models.Model):
     _name = 'account.invoice'
     _inherit = ['account.invoice', 'base.ubl']
 
-    # TODO: maybe we will have to switch from UBL 2.1 to UBL 2.0...
-    # to be e-fff compliant
     @api.multi
-    def _ubl_add_header(self, parent_node, ns):
+    def get_ubl_version(self):
+        '''e-fff is based on UBL 2.0, not UBL 2.1'''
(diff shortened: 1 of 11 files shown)


l10n_be_invoice_ubl/models/account_invoice.py
```diff
@@ -0,0 +1,45 @@
...
+        parent_node.insert(
+            parent_node.index(version_node_find) + 1, customization_id)
+        profile_id = etree.Element(ns['cbc'] + 'ProfileID')
+        profile_id.text = 'EFFF.BE Odoo %s' % release.version
+        parent_node.insert(
+            parent_node.index(version_node_find) + 2, profile_id)
+        return res
+
+    @api.multi
+    def get_ubl_filename(self, version='2.1'):
```
houssine78, line 38: should it be version 2.0?

l10n_be_invoice_ubl/models/account_invoice.py
```diff
@@ -0,0 +1,45 @@
...
+        '''Read from http://www.e-fff.be/FR/doc03.php :
+        The e-fff community recommends to use the following file naming:
+        efff_BE0123456789_AlphaNumericCharactersFreeOfChoice.xml'''
+        filename = 'efff_%s_Invoice_%s' % (
+            self.commercial_partner_id.sanitized_vat,
+            self.number and self.number.replace('/', '-') or 'draft')
+        return filename
```
houssine78, line 45: add +'.xml'

l10n_be_invoice_ubl/models/account_invoice.py
alexis-via: This is the proto of the method that we inherit, so we don't change it in the proto.

l10n_be_invoice_ubl/models/account_invoice.py
alexis-via: Good catch !

rousseldenis:
For information, found Peppol website that defines Pan European specifications (UBL)
http://www.peppol.eu/ressource-library/technical-specifications/post-award/mandatory

houssine78:
+1

alexis-via:
Is e-fff so unpopular in Belgium that nobody's interested in this PR ?

@sbidoul I don't see any field set as "Mandatory" in the specs which is not present in the UBL file generated by Odoo (given that the appropriate info is set in Odoo). Could you tell us which field you see as mandatory in the spec that is not set by Odoo ?

sbidoul:
@alexis-via it could be the intersection of the set of e-fff users and the set of odoo users is empty.

We may change that this year though. Patience.

Regarding the mandatory fields, I don't know yet. That todo was yours :)

sbidoul:
See also https://github.com/odoo/odoo/pull/17278

PR #38, opened 2016-11-15
[9.0][ADD] l10n_be_apb_tax: APB taxe

jbaudoux:
Implements the belgian APB tax.

Tax on drugs
2,24c€ / unit
not subjected to VAT
only for sales
Applicable to national clients excepted wholesalers (a fiscal position is created to exempt them of the tax)


sbidoul (changes requested):
@jbaudoux thanks for this contrib!

Can you check the README matches the latest OCA template? https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

Also the .pot file is unrelated. You can remove it as it's not necessary to translate on transifex.

Otherwise looks good to me.

jbaudoux:
@sbidoul Thanks for the review. I solved readme and pot.

PR #39, opened 2016-11-15
[9.0][ADD] l10n_be_antibiotic_tax: antibiotics taxes

jbaudoux:
Tax on antibiotics.


sbidoul (approved):
It would be gread if you could check the README is aligned with the latest OCA template. https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

Otherwise looks good.

jbaudoux:
@sbidoul Thanks for the review. I solved readme.

PR #50, opened 2017-10-21
[IMP] l10n_be_mis_reports: add new EUR style

sbidoul:


lmignon (approved):
Code review

vdewulf:
Tested OK on v10.0 and on v9.0 (after back-porting).
The templates data have now 2 decimals instead of 0.

PR #53, opened 2017-10-27
[9.0] Mis builder Report : Belgian VAT Declaration

vdewulf:
This PR adds a new template in the MIS Builder Templates for Belgium : the VAT Declaration. It is based on the tags on the taxes. The translation has been made in French only.
Improvement to come : add the series of logical and arithmetical checks that are usually provided by accounting tools.

It has been validated with an accounting expert for the more usual grids (01, 03, 49, 81, 82, 85, 54, 63, 59, 64, 71/72).

@sbidoul here is my contribution for the VAT report

### DIFF
--- a/l10n_be_mis_reports/__openerp__.py
+++ b/l10n_be_mis_reports/__openerp__.py
@@ -39,6 +39,7 @@
         'data/mis_report_styles.xml',
         'data/mis_report_pl.xml',
         'data/mis_report_bs.xml',
+        'data/mis_report_vat.xml',
     ],
     'installable': True,
 }
--- a/l10n_be_mis_reports/__openerp__.py
+++ b/l10n_be_mis_reports/__openerp__.py
@@ -34,6 +34,7 @@
     'license': 'AGPL-3',
     'depends': [
         'mis_builder',  # OCA/account-financial-reporting
+        'l10n_be',
     ],
     'data': [
         'data/mis_report_styles.xml',
--- a/l10n_be_mis_reports/data/mis_report_vat.xml
+++ b/l10n_be_mis_reports/data/mis_report_vat.xml
@@ -22,7 +22,7 @@
       <field name="name">cadre_2</field>
       <field name="description">Cadre II</field>
       <field name="expression"></field>
-      <field name="type">num</field>
+      <field name="type">str</field>
       <field name="compare_method">pct</field>
       <field name="style_id" ref="mis_report_style_l10n_be_2"/>
       <field name="sequence">1</field>
@@ -31,7 +31,7 @@
       <field name="report_id" ref="mis_report_vat"/>
       <field name="name">grid_00</field>
       <field name="description">Grid 00</field>
-      <field name="expression">crd[][('tax_ids.tag_ids.name', '=', 'Belgium VAT Form: grid 0')]</field>
+      <field name="expression">crd[][('tax_ids.tag_ids.id', '=', ref('l10n_be.tax_tag_00').id)]</field>
       <field name="type">num</field>
       <field name="compare_method">pct</field>
       <field name="style_id" ref="mis_report_style_l10n_be_3"/>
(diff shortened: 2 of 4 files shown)


l10n_be_mis_reports/data/mis_report_vat.xml
```diff
@@ -0,0 +1,369 @@
...
+      <field name="type">num</field>
+      <field name="compare_method">pct</field>
+      <field name="style_id" ref="mis_report_style_l10n_be_3"/>
+      <field name="sequence">80</field>
+    </record>
+    <record model="mis.report.kpi" id="mis_report_vat_grid_48">
+      <field name="report_id" ref="mis_report_vat"/>
+      <field name="name">grid_48</field>
+      <field name="description">Grid 48</field>
+      <field name="expression">bal[][('tax_ids.tag_ids.name', 'like', '% 48'),('debit','>',0)] </field>
```
sbidoul: For cases like this where we want only debits, it's better to use `deb[][('tax_ids.tag_ids.name', 'like', '% 48')]`. Performance will be better.

vdewulf:
@sbidoul Thanks a lot for the review. I replaced all the "bal" expressions with "deb" and "crd" according to your advice.

sbidoul:
@vdewulf I created https://github.com/vdewulf/l10n-belgium/pull/1 to add your report as a menu.

l10n_be_mis_reports/__openerp__.py
```diff
@@ -39,6 +39,7 @@
         'data/mis_report_styles.xml',
         'data/mis_report_pl.xml',
         'data/mis_report_bs.xml',
+        'data/mis_report_vat.xml',
     ],
     'installable': True,
 }
```
sbidoul, line 46: Now that we depends on the menu (l10n_be.account_reports_be_statements_menu) we need to add the dependency on l10n_be here.

vdewulf:
@sbidoul I added the dependency as suggested.
Thanks!

l10n_be_mis_reports/data/mis_report_vat.xml
```diff
@@ -0,0 +1,472 @@
...
+      <field name="style_id" ref="mis_report_style_l10n_be_3"/>
+      <field name="sequence">904</field>
+    </record>
+     <record model="mis.report.kpi" id="mis_report_vat_control_87">
+      <field name="report_id" ref="mis_report_vat"/>
+      <field name="name">ctrl_87</field>
+      <field name="description">Control Grid 87: [85] + [87] x 0.21 >= [56] + [57]</field>
+      <field name="expression">min((grid_85 + grid_87)*0.21-grid_57-grid_56,AccountingNone)</field>
+      <field name="type">num</field>
+      <field name="compare_method">pct</field>
```
sbidoul: Very very nice you add these checks.

Would you like to try this kind of expression (using type=str) ?

`'✔' if (grid_85 + grid_87)*0.21 >= (grid_57 + grid_56) else '✘'`

Or do you think having the actual difference in the cell is better?

l10n_be_mis_reports/data/mis_report_vat.xml
```diff
@@ -0,0 +1,472 @@
...
+      <field name="style_id" ref="mis_report_style_l10n_be_3"/>
+      <field name="sequence">280</field>
+    </record>
+    <record model="mis.report.kpi" id="mis_report_vat_grid_64">
+      <field name="report_id" ref="mis_report_vat"/>
+      <field name="name">grid_64</field>
+      <field name="description">Grid 64</field>
+      <field name="expression">deb[][('tax_line_id.tag_ids.name', 'like', '% 64')]</field>
```
sbidoul: Now that mis_builder version 3.1 is merged, it is possible to write

`deb[][('tax_line_id.tag_ids.id', '=', ref('l10n_be.tax_tag_64').id)]`

This would be more robust in case someone decides to translate or otherwise change the tag labels. 

Not blocking for me though.

vdewulf:
@sbidoul I validated the controls with an accounting expert.
In principles, my work is done on this topic.
Could you consider merging this?
Thanks a lot!

sbidoul:
@vdewulf thanks a lot for this fantastic work.

In principle we should need another review but since we are a small population here, I'll merge.

Note I'll soon work on v10 and the intracom declaration.

PR #54, opened 2017-12-05
[10.0] mig bank statement import coda from 8.0

sbidoul:


sbidoul:
@Olivier-LAURENT have you had a chance to test this?

Olivier-LAURENT (approved):
Reviewed and tested, thx for the migration. 👍

account_bank_statement_import_coda/wizard/account_bank_statement_import_coda.py
```diff
@@ -0,0 +1,196 @@
...
+            raise UserError(
+                _("The following problem occurred during import. "
+                  "The file might not be valid.\n\n %s" % e.message))
+
+        acc_number = None
+        currency = None
+        if statements:
+            acc_number = statements[0].acc_number
+            currency = statement.currency
+        return currency, acc_number, vals_bank_statements
```
houssine78, line 62: in our version we return  return currency, self._get_acc_number(account_number), vals_bank_statements

this _get_acc_number function ensure that we have a defined journal for this account number.

    def _get_acc_number(self, acc_number):
        #Check if we match the exact acc_number or the end of an acc number
        journal = self.env['account.journal'].search([('bank_acc_number', '=like', '%' + acc_number)])
        if not journal or len(journal) > 1: #if not found or ambiguious 
            return acc_number
        
return journal.bank_acc_number

account_bank_statement_import_coda/wizard/account_bank_statement_import_coda.py
```diff
@@ -0,0 +1,196 @@
...
+            }
+        """
+        balance_start = statement.old_balance
+        if statement.old_balance_amount_sign == AmountSign.DEBIT:
+            balance_start = - balance_start
+        balance_end_real = statement.new_balance
+        if statement.new_balance_amount_sign == AmountSign.DEBIT:
+            balance_end_real = - balance_end_real
+        transactions = []
+        statement_date = statement.new_balance_date
```
houssine78, line 84: why not using statement.creation_date?

houssine78 (commented):
@sbidoul I've just two small remarks which will be great for us to be adressed. Otherwise it could be merged.

vdewulf:
Tested NOT OK with Triodos bank.
I imported a first CODA without any problem, with 6 transactions.
Then I tried to import the next CODA file. The system told me that the file was already imported. 
I tried with another CODA file, it was partially imported (the system told me that some transactions were ignored, in fact the 6 first).

It seems that the unique reference construction should be improved. See houssine's feedback for the technical details.
Here are an extract of two CODA files that I tried to import:
image

account_bank_statement_import_coda/wizard/account_bank_statement_import_coda.py
```diff
@@ -0,0 +1,196 @@
...
+            amount = - amount
+        return {
+            'name': self.get_st_line_name(line, globalisation_dict),
+            'date': line.entry_date or datetime.datetime.now().date(),
+            'amount': amount,
+            'ref': line.ref,
+            'partner_name': line.counterparty_name or None,
+            'account_number': line.counterparty_number or None,
+            'note': self.get_st_line_note(line, information_dict),
+            'unique_import_id': line.ref + line.transaction_ref,
```
houssine78, line 186: we get a problem here testing it with Triodos Coda.
the uniqe_import_id is too weak and led to ignore statement and or statement line import.
in our implementation we build the uniqe_import_id  like this statement.coda_seq_number + '-' + statement.old_balance_date + '-' + statement.new_balance_date + '-' + move.ref

we also found that trasanction_ref returned from pycoda is in fact the bank reference from which the payment as been made. This reference is not mandatory and can be empty(N° de référence de la banque. Cette information est strictement informative)

vdewulf:
@sbidoul Hello Stéphane. Is there any news regarding our test results on this PR?
Thanks in advance!

vdewulf:
@sbidoul Hi Stéphane,
Is it possible for you to merge (even if the error we found is not fixed yet)?
So that we can migrate to v12 and propose a fix in another PR.
Thanks for your follow-up!
@houssine78 for information

sbidoul:
@vdewulf I've backported the most changes from v11. Let's see if it's green.

/ocabot merge

houssine78:
thanks @sbidoul

PR #55, opened 2017-12-08
[10.0] migrate VAT report

sbidoul:
cc/ @vdewulf


vdewulf:
Tested on runbot. Everything looks fine and the icons (✔ and ✘) are a nice improvement!
The best control would be to test this module with real data from a customer and compare the results with a VAT report sent to the administration. I have no such customer on v10 unfortunately.

sbidoul:
Ok, I merge since it's a straightforward port.

PR #57, opened 2018-02-01
Reports ported: VAT Listing and VAT Intra

andreparames:
This is a straight migration of two reports to version 10.

### DIFF
--- a/l10n_be_vat_reports/wizard/l10n_be_partner_vat_listing.py
+++ b/l10n_be_vat_reports/wizard/l10n_be_partner_vat_listing.py
@@ -82,8 +82,7 @@ WITH turnover_tags AS
        AND tagsrel.account_account_tag_id = tag_xmlid.res_id)
    WHERE tag_xmlid.NAME IN ('tax_tag_54',
                             'tax_tag_64'))
-SELECT sub1.partner_id,
-       sub1.NAME,
+SELECT sub1.NAME,
        sub1.vat,
        COALESCE(sub1.turnover, 0) AS turnover,
        COALESCE(sub2.vat_amount, 0) AS vat_amount
@@ -361,9 +360,9 @@ class PartnerVATList(models.TransientModel):
 
 class PartnerVATListingPrint(report_sxw.rml_parse):
 
-    @api.model
-    def __init__(self, name):
-        super(PartnerVATListingPrint, self).__init__(name)
+    def __init__(self, cr, uid, name, context=None):
+        super(PartnerVATListingPrint, self).__init__(
+            cr, uid, name, context=context)
         self.localcontext.update({
             'time': time,
         })
--- a/l10n_be_vat_reports/wizard/l10n_be_partner_vat_listing.py
+++ b/l10n_be_vat_reports/wizard/l10n_be_partner_vat_listing.py
@@ -212,7 +212,7 @@ class PartnerVATList(models.TransientModel):
         obj_sequence = self.env['ir.sequence']
         obj_partner = self.env['res.partner']
         obj_model_data = self.env['ir.model.data']
-        seq_declarantnum = obj_sequence.get('declarantnum')
+        seq_declarantnum = obj_sequence.next_by_code('declarantnum')
         obj_cmpny = self.env.user.company_id
         company_vat = obj_cmpny.partner_id.vat
 
@@ -222,7 +222,7 @@ class PartnerVATList(models.TransientModel):
         company_vat = company_vat.replace(' ', '').upper()
         SenderId = company_vat[2:]
         issued_by = company_vat[:2]
-        seq_declarantnum = obj_sequence.get('declarantnum')
+        seq_declarantnum = obj_sequence.next_by_code('declarantnum')
         dnum = company_vat[2:] + seq_declarantnum[-4:]
         street = city = country = ''
         addr = obj_cmpny.partner_id.address_get(['invoice'])
--- a/l10n_be_vat_reports/wizard/l10n_be_partner_vat_listing.py
+++ b/l10n_be_vat_reports/wizard/l10n_be_partner_vat_listing.py
@@ -362,13 +362,6 @@ class PartnerVATList(models.TransientModel):
 
 class PartnerVATListingPrint(report_sxw.rml_parse):
 
-    def __init__(self, cr, uid, name, context=None):
-        super(PartnerVATListingPrint, self).__init__(
-            cr, uid, name, context=context)
-        self.localcontext.update({
-            'time': time,
-        })
-
     def set_context(self, objects, data, ids, report_type=None):
         client_datas = data['client_datas']
         self.localcontext.update({
--- a/l10n_be_vat_reports/wizard/l10n_be_partner_vat_listing.py
+++ b/l10n_be_vat_reports/wizard/l10n_be_partner_vat_listing.py
@@ -1,0 +1,371 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    Copyright (c) 2018 ACSONE SA/NV (<http://acsone.eu>)
+#    Copyright (C) 2004-2010 Tiny SPRL (<http://tiny.be>).
+#
+#    Corrections & modifications by Noviat nv/sa, (http://www.noviat.be):
+#    - VAT listing based upon year in stead of fiscal year
+#    - sql query adapted to select only 'tax-out' move lines
+#    - extra button to print readable PDF report
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+import time
+import base64
+
+from openerp import fields, models, api, _
+from openerp.report import report_sxw
+from openerp.exceptions import Warning
+
+
+class VATListingClients(models.TransientModel):
+    _name = 'vat.listing.clients'
+
+    name = fields.Char('Client Name')
+    vat = fields.Char('VAT')
+    turnover = fields.Float('Base Amount')
+    vat_amount = fields.Float('VAT Amount')
+
+
+class PartnerVAT(models.TransientModel):
+    """ Vat Listing """
+    _name = "partner.vat"
+
+    @api.multi
+    def get_partner(self):
+        context = dict(self.env.context)
+        obj_partner = self.env['res.partner']
+        obj_vat_lclient = self.env['vat.listing.clients']
+        obj_model_data = self.env['ir.model.data']
+        data = self.read()[0]
+        year = data['year']
+        date_start = year + '-01-01'
+        date_stop = year + '-12-31'
+
+        partners = obj_vat_lclient.browse([])
+        be_partners = obj_partner.search([('vat', 'ilike', 'BE%')])
+        if not be_partners:
+            raise Warning(_('No belgium contact with a VAT number '
+                            'in your database.'))
+        query = """
+            SELECT sub1.partner_id, sub1.name, sub1.vat,
+                   COALESCE(sub1.turnover, 0) as turnover,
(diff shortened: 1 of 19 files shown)


l10n_be_vat_reports/wizard/l10n_be_partner_vat_listing.py
```diff
@@ -0,0 +1,371 @@
...
+                      ON t.id = tagsrel.account_tax_id
+                  LEFT JOIN ir_model_data tag_xmlid ON (
+                      tag_xmlid.model = 'account.account.tag'
+                      AND tagsrel.account_account_tag_id = tag_xmlid.res_id)
+                  WHERE tag_xmlid.name IN (
+                      'tax_tag_00', 'tax_tag_01', 'tax_tag_02',
+                      'tax_tag_03', 'tax_tag_45', 'tax_tag_49')
+                  AND l.partner_id IN %s
+                  AND l.date BETWEEN %s AND %s
+                  GROUP BY l.partner_id, p.name, p.vat) AS sub1
```
sbidoul: @andreparames the turnover computation is incorrect. 

In v9+, we must sum move lines where "tax_ids has one of (00,01,02,03,45,49)" to find the tax bases.

andreparames: But that's what I do (in lines 78-83), it's just that I search using the XML-IDs. I find all the account_tax with a tag that has on of the following XML-IDs: 'tax_tag_00', 'tax_tag_01', 'tax_tag_02', 'tax_tag_03', 'tax_tag_45', 'tax_tag_49', which are defined in the v9 of l10n_be: https://github.com/odoo/odoo/blob/9.0/addons/l10n_be/account_tax_template.xml#L5

Is this incorrect? Should I be hardcoding the account_tax ids?

andreparames:
@sbidoul thanks for the heads up! I've pushed a fix for that bug and another which I discovered during testing.

sbidoul:
As discussed:
- use  `-balance` for turnover and tax
- replace partition by some sort of where exists (btw, the list of tax tags ids could be prefetched with the ORM)
- there is probably a similar change to do for the vat_intra report
- a couple of line too long errors

sbidoul:
Or even better, `search()` for the taxes with the corresponding tax and then use `tax_line_id in %s` and `where exists` on the tax_ids relation.

andreparames:
I wrote them as CTEs, just tell me if you prefer the `search()` solution anyway.

I've also added a test for the VAT listing that checks one value in the resulting XML.

vdewulf:
Hello,
I tested locally the Annual Listing of VAT-Customer on version 9 (I just replaced the name of the file "__manifest__" by "__openerp__".
When there is no VAT number, it doesn't appear in the list, which I suppose is the intended behavior).

# Tested OK
One customer has 2 invoices. Each invoice is about 100€ hvat + 21% vat.
=> it doesn't appear on the list. OK
I created a 3th invoice (100€ hvat). When generating the Annual Listing wizard, there is a line with 300€ and 63€ vat and the corresponding VAT number.
=> OK

I also tested the limit amount by changing its value. The corresponding listing is correctly updated.

# To confirm
When there is no tax set on the invoice line, the corresponding amount does not appear in the total of the Base Amount.
=> Is it intended? (I'm thinking about the "régime de la franchise de TVA" where the company has to give such listing even if no VAT was invoiced to the customer). Should be confirmed by an accountant.

I tested the XML creation. The XML is created but I don't know the requirements to go futher on this testing.

# Tested Not OK
I tested the printing of the listing to a PDF. I got an error (AttributeError: 'list' object has no attribute 'ids').
Details of the log in attachment.
[Error_Printing_VAT_Listing_to_PDF.txt](https://github.com/OCA/l10n-belgium/files/1866489/Error_Printing_VAT_Listing_to_PDF.txt)

andreparames:
@vdewulf thanks for the test. The behavior in v10 was made more robust in Odoo, but I've pushed a change that should make it work on both versions. Can you try on v9 again?

vdewulf:
@andreparames I tested with the last version on v9 as proposed.
## Tested OK ##
I tested the printing of the listing to a PDF ant it worked this time.
I re-tested changing the value in Limit Amount and the result was correctly updated.
The refund invoices are well taken into account in the total amount in the list.
I compared the result with a listing made my an accoutant for a customer and the result is the same (except some rounding differences).

## Confirmed (cf. my previous comment) ##
For my question about the special "régime de franchise TVA", on the invoice, the tax 0% should be used so that these amount would appear in the listing (I suppose it's the case, I did not tested this specific case).

## Tested Not OK ##
I tried to generate the XML but a pop up appeared with the message "No email address associated with the company.".
However, I have an email set in the general settings of the company.

In the logs, following messages appeared when I clicked on the button to create the XML file:
`2018-04-11` 13:23:45,620 11969 WARNING medor-vat-listing openerp.addons.base.ir.ir_sequence: ir_sequence.get() and ir_sequence.get_id() are deprecated. Please use ir_sequence.next_by_code() or ir_sequence.next_by_id().
`2018-04-11` 13:23:45,635 11969 WARNING medor-vat-listing openerp.addons.base.ir.ir_sequence: ir_sequence.get() and ir_sequence.get_id() are deprecated. Please use ir_sequence.next_by_code() or ir_sequence.next_by_id().

Maybe it's due to my DB, bad configuration... ? I didn't test on another DB.

andreparames:
@vdewulf the email address must be on the Invoicing address of the company. I'm not sure if this still makes sense, but it was the original behavior. If you add one, it should work.

I'll fix those warnings, thanks for pointing them out!

vdewulf:
@andreparames Indeed, it worked. I had a message about the phone that was missing, I put one on the invoicing adress and then the XML could be generated.

In the XML, there are encoding issues (for example for the city of Liège: <City>LiÃ¨ge</City>)
I don't know if this should be fixed or not (does the administration accept this sort of file?).
For information I only tested the Annual Listing of VAT-Customer, not the other report.

Another improvement: If MIS Builder is installed together with the Belgian Templates (Belgium P&L, Balance Sheet, VAT Declaration), there is in the menu "Report" a sub-menu named "Belgium Statements" under which there is "VAT Declaration".
With this module l10n_be_vat_reports installed, there is another sub-menu named "Belfium Statements" under which there are the VAT Listing and the Intra Comm report.
Would it be possible to merge the submenus in one only?
See screenshot attached:
[Menu_Belgium_Statements.pdf](https://github.com/OCA/l10n-belgium/files/1899104/Menu_Belgium_Statements.pdf)

Thanks a lot!

l10n_be_vat_reports/wizard/l10n_be_partner_vat_listing.xml
```diff
@@ -0,0 +1,118 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<openerp>
+  <data>
+
+    <menuitem
+        id="menu_finance_belgian_statement"
+        name="Belgium Statements"
+        parent="account.menu_finance_reports" groups="account.group_account_manager"/>
```
sbidoul: This menu already exists in l10n_be indeed.

See example use here: https://github.com/OCA/l10n-belgium/blob/10.0/l10n_be_mis_reports/data/mis_report_vat.xml#L12

sbidoul:
> In the XML, there are encoding issues (for example for the city of Liège: LiÃ¨ge)

@vdewulf Can you try opening the xml file in firefox? Or attach here so we can have a look.

vdewulf:
@sbidoul Here is a screenshot after opening in Firefox. As these are real date from our customer, I prefer not to put the whole file here.
xml_open_in_firefox

l10n_be_vat_reports/wizard/l10n_be_partner_vat_listing.py
```diff
@@ -0,0 +1,388 @@
...
+
+        data_end = """
+
+        <ns2:Comment>%(comments)s</ns2:Comment>
+    </ns2:ClientListing>
+</ns2:ClientListingConsignment>
+""" % annual_listing_data
+
+        data_file += data_begin + data_comp + data_client_info + data_end
+        file_save = base64.encodestring(data_file.encode('utf8'))
```
sbidoul, line 328: @andreparames here encode('ISO-8859-1') instead of utf8 as that's what indicated in the XML declaration above.

l10n_be_vat_reports/wizard/l10n_be_vat_intra.py
```diff
@@ -0,0 +1,376 @@
...
+            '\n\t<ns2:IntraListing SequenceNumber="1" '
+            'ClientsNbr="%(clientnbr)s" DeclarantReference="%(dnum)s" '
+            'AmountSum="%(amountsum).2f">') % (xml_data)
+
+        data_file += (
+            data_head + data_decl + data_comp_period + data_clientinfo +
+            '\n\t\t<ns2:Comment>%(comments)s</ns2:Comment>\n\t'
+            '</ns2:IntraListing>\n</ns2:IntraConsignment>') % (xml_data)
+        context = dict(self.env.context)
+        context['file_save'] = data_file
```
sbidoul: encode ISO-8859-1 too here?

l10n_be_vat_reports/wizard/l10n_be_partner_vat_listing.py
andreparames: While this is clearly problematic, it was already outputting in UTF-8, so isn't it risky to change now? Should we change the declaration instead?

l10n_be_vat_reports/wizard/l10n_be_partner_vat_listing.xml
andreparames: Thanks for the warning!

l10n_be_vat_reports/wizard/l10n_be_partner_vat_listing.py
sbidoul: Ok, let's try with changing the declaration

andreparames:
Ok, I've merged the two menus and have updated the XML declarations.

l10n_be_vat_reports/wizard/l10n_be_vat_intra.py
```diff
@@ -116,7 +109,7 @@ def _get_datas(self):
         if len(wiz_data.period_code) != 6:
             raise UserError(_('Period code is not valid.'))
 
-        if not wiz_data.date_start > wiz_data.date_end:
+        if not wiz_data.date_start <= wiz_data.date_end:
```
sbidoul, line 112: remove the `not`

andreparames:
Thanks for the review. I've pushed some fixes to the Intra report, plus a test file.

vdewulf:
@andreparames Hi! Do you want me to test the last version? Just tell me! Thanks.

andreparames:
@vdewulf sure, the latest version should fix the two problems you identified (file encoding and duplicated menu).

vdewulf:
@andreparames 
### VAT listing
I tested OK the encoding fix and the unified menu. Seems OK for the VAT listing.
### Intracomm report
I tested the intracomm report. When I click on "preview" button, it raised an error :
`File "/home/virg/odoo-dev/odoo/addons/report/models/report.py", line 307, in get_action
    return Report.get_action(self._model, self._cr, self._uid, records.ids,`
`AttributeError: 'list' object has no attribute 'ids'`
while an XML is being created.
In the XML file, there is a False value in this code: <ns2:Code>False</ns2:Code>
Is it intended? What should be there?
intracomm

andreparames:
@vdewulf thanks for the review, I pushed an update that should fix the remaining problems.

vdewulf:
I came back to this PR. sorry for the delay!

VAT Listing
I retested the basic scenario, testing when there is no VAT number on a partner (it does not appaear in the list), when the limit is higher or lower than 250€ (the corresponding info are displayed).
I tested the printing, it worked fine.
I tested the XML generation, it worked fine.

Intracomm report
I tested the printing and the XML generation.
The False value detected previously has disapeared (I got S code instead).

I think that works fine, but I would like to double check with more data and and expert in accounting.

Okay for me, thanks for the work!

sbidoul:
Ok. Before merging we'll need to do some rework of the commits, to have a first commit that is exactly identical to the Odoo 8 version, and then add our changes on top, so copyright can be properly attributed to the part that comes from Odoo SA.

vdewulf:
@sbidoul Hello Stephane. Happy New Year 2019!
I'd like to know if this PR will soon be merged or not?
It will be need by our clients in march 2019 in v9.
Thanks in advance!

sbidoul:
Hi @vdewulf my previous comment must be done by someone before merging. It's important to avoid copyright issues with Odoo SA. Can one of your colleagues do it?

vdewulf:
Hi @sbidoul. I'll check with my colleagues!

robinkeunen:
Hi @sbidoul 

Is this something like #74 you were expecting?

sbidoul:
Closing in favor of #74

PR #64, opened 2018-07-04
[FIX] vat report: add rounding in checks and other fixes and improvements

sbidoul:
Without these, we can have false positives.

@vdewulf I'd love to get your review here :)

### DIFF
--- a/l10n_be_mis_reports/data/mis_report_vat.xml
+++ b/l10n_be_mis_reports/data/mis_report_vat.xml
@@ -324,7 +324,7 @@
       <field name="report_id" ref="mis_report_vat"/>
       <field name="name">grid_62</field>
       <field name="description">Grid 62</field>
-      <field name="expression">crd[][('tax_line_id.tag_ids.id', '=',ref('l10n_be.tax_tag_62').id)]</field>
+      <field name="expression">deb[][('tax_line_id.tag_ids.id', '=',ref('l10n_be.tax_tag_62').id)]</field>
       <field name="type">num</field>
       <field name="compare_method">pct</field>
       <field name="style_id" ref="mis_report_style_l10n_be_3"/>
--- a/l10n_be_mis_reports/data/mis_report_vat.xml
+++ b/l10n_be_mis_reports/data/mis_report_vat.xml
@@ -234,7 +234,7 @@
       <field name="report_id" ref="mis_report_vat"/>
       <field name="name">grid_54</field>
       <field name="description">Grid 54</field>
-      <field name="expression">crd[][('tax_line_id.tag_ids.id', '=', ref('l10n_be.tax_tag_54').id)] </field>
+      <field name="expression">-bal[][('tax_line_id.tag_ids.id', '=', ref('l10n_be.tax_tag_54').id), ('debit', '=', 0)]</field>
       <field name="type">num</field>
       <field name="compare_method">pct</field>
       <field name="style_id" ref="mis_report_style_l10n_be_3"/>
@@ -244,7 +244,7 @@
       <field name="report_id" ref="mis_report_vat"/>
       <field name="name">grid_55</field>
       <field name="description">Grid 55</field>
-      <field name="expression">deb[][('tax_line_id.tag_ids.id', '=', ref('l10n_be.tax_tag_55').id)]</field>
+      <field name="expression">bal[][('tax_line_id.tag_ids.id', '=', ref('l10n_be.tax_tag_55').id), ('credit', '=', 0)]</field>
       <field name="type">num</field>
       <field name="compare_method">pct</field>
       <field name="style_id" ref="mis_report_style_l10n_be_3"/>
@@ -254,7 +254,7 @@
       <field name="report_id" ref="mis_report_vat"/>
       <field name="name">grid_56</field>
       <field name="description">Grid 56</field>
-      <field name="expression">deb[][('tax_line_id.tag_ids.id', '=', ref('l10n_be.tax_tag_56').id)]</field>
+      <field name="expression">bal[][('tax_line_id.tag_ids.id', '=', ref('l10n_be.tax_tag_56').id), ('credit', '=', 0)]</field>
       <field name="type">num</field>
       <field name="compare_method">pct</field>
       <field name="style_id" ref="mis_report_style_l10n_be_3"/>
@@ -264,7 +264,7 @@
       <field name="report_id" ref="mis_report_vat"/>
       <field name="name">grid_57</field>
       <field name="description">Grid 57</field>
-      <field name="expression">deb[][('tax_line_id.tag_ids.id', '=', ref('l10n_be.tax_tag_57').id)]</field>
+      <field name="expression">bal[][('tax_line_id.tag_ids.id', '=', ref('l10n_be.tax_tag_57').id), ('credit', '=', 0)]</field>
       <field name="type">num</field>
       <field name="compare_method">pct</field>
       <field name="style_id" ref="mis_report_style_l10n_be_3"/>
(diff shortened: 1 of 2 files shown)


vdewulf:
Tested today on v9.0 on real data (for one organization for the Q2 VAT report).
I compared a VAT report without this PR and with this PR (see attachment).

I actually had false negatives without this PR for two checks (55, 87), that were identified as not correct "X" but actually was "0" = "0". With this rounding added, the "X" became "V".

For the check on 54, there are still rounding issues in this specific test, with a result of "X" while there is only 1 cent of difference. I don't know if this was expected to be solved with this PR.
selection_715

Hope this helps!

vdewulf (approved):
See my comment with attachment.

sbidoul:
@vdewulf I made a change to the check on 54, can you try again?

vdewulf:
Tested with the same data but the issue is still there.
Actually, there are only values in the grid [03], so the calculation is grid[03] * 0.21 - grid[54], so the commit does not change anything on this specific issue.
The data are : 14393.56 in grid [03]
=> 14393.56 * 0.21 = 3022.6476
=> rounded : 3022,65
Grid[54] displays 3022.66

sbidoul:
Do you know if there is a tolerance in the official checks?

sbidoul:
https://eservices.minfin.fgov.be/intervat/static/help/FR/helpfile.htm#regles_de_validation_d_une_declaration.htm

sbidoul:
@vdewulf I dig out the intervat validation rules, this should be better. 

But we must review the other checks too.

vdewulf:
Hello,
I tested the last updated check on grid 54. In case there are no values in grid 01 or 02 (in my test case), the result is still "x".
image

In this case:
None + None + 0.2
Traceback (most recent call last):
  File "<input>", line 1, in <module>
TypeError: unsupported operand type(s) for +: 'NoneType' and 'NoneType'

For the other changes (rounding to 2 decimals), I already tested OK previously.

For the other checks for Intervat, I can try and add them. I will need support to add the conditions though (for example : if there is a value in grid xx, then there must be ...).

vdewulf:
Hello,
I tested the last updated check on grid 54.
image
It works fine it seems, even when there is no value in grid 01 nor grid 02 at all.

But I got errors for the checks on 55 and 87. For these checks, there are no values at all in the corresponding grids.
I changed the check on grid 55 (implementing check P from intervat) by
`u'✔' if abs((grid_84 + grid_86 + grid_88)*0.21 - grid_55) &lt; 62 else u'✘'`
 and the check on grid 87 (implementing check Q from intervat) by
`u'✔' if abs((grid_85 + grid_87)*0.21 - (grid_56 + grid_57)) &lt; 62 else u'✘'`

==> it worked fine:
image

(I don't know if I can commit and push on this PR. I let you do it?!).

For the other changes (rounding to 2 decimals), I already tested OK previously.

For the other checks for Intervat, I can try and add them. I will need support to add the conditions though (for example : if there is a value in grid xx, then there must be ...). Shall we continue on this PR or open a new one?

sbidoul:
The best is you do a PR to this branch. 

Note I also made https://github.com/OCA/l10n-belgium/pull/64/commits/e6cf09caf8f6f8fbc25604107a6c65b69aa53fe6 to be more precise when deciding about invoice/credit note.

Let's accumulate all this improvements here. Then I'll forward port to 10 and 11.

Also make sure to use the very latest version of mis builder which may fix some errors you see.

sbidoul:
Note e6cf09c is not fully finished. I think the same mechanism must be applied to some other grids as well.

sbidoul:
@vdewulf note some intervat checks are not needed here (eg those which simply checks grids that are sums of other grids): they will always be correct as they are computed by our formulas.

vdewulf:
I tested the last commit (with the latest version of mis builder as well) about invoice/credit note.
I got this error when generating a VAT report:
`  File "/home/virg/odoo-dev/odoo/openerp/tools/safe_eval.py", line 153, in test_expr
    code_obj = compile(expr, "", mode)
  File "<string>", line 1
    [('tax_ids.tag_ids.id', 'in', (ref('l10n_be.tax_tag_46L').id), ref('l10n_be.tax_tag_46T').id)), '|', ('invoice_id.type', '=', 'out_invoice'), ('debit', '=', 0)]
                                                                                                 ^
SyntaxError: invalid syntax`

I cannot see where the syntax problem is.

sbidoul:
Sorry about that. Here is the good one
```
 [('tax_ids.tag_ids.id', 'in', (ref('l10n_be.tax_tag_46L').id), ref('l10n_be.tax_tag_46T').id)), '|', ('invoice_id.type', '=', 'out_invoice'), ('debit', '=', 0)]
```

sbidoul:
Sorry about that, here is the good one (bad closing parenthesis after the first `ref`
```
[('tax_ids.tag_ids.id', 'in', (ref('l10n_be.tax_tag_46L').id, ref('l10n_be.tax_tag_46T').id)), '|', ('invoice_id.type', '=', 'out_invoice'), ('debit', '=', 0)]
```

vdewulf:
Hi,
I reviewed all the changes made on the grids in this PR.

### Grid 00/01/02/03/44/45/46/47
I have doubt concerning the changes made.
Without the changes, the system takes into account only the positive amounts, without the negatives amount.
With the changes, the system takes into account positive and negative amounts, and reduced the total amount by doing so. This does not seem to be compliant with the administration's rules (cf. attachment - See the articles 98, 99, 100^5, 100^6, 108, 109, 120, 121, 131, 132). Page 50 gives a summary of all these rules.

Here is an abstract in French, page 31 :
"Notes de crédit et corrections négatives ou positives
98. Le montant (hors TVA) des notes de crédit délivrées au cours de la périodeà laquelle se rapporte la déclaration pour des opérations inscrites dans les grilles 01, 02 ou 03 de la déclaration ou d'une déclaration antérieure, ne peut pas être déduit du montant inscrit dans les grilles 01, 02 ou 03. Il doit être mentionné distinctement en grille 49 (v. n° 141).
99. Le même principe est appliqué aux corrections négatives.

Moreover, I know some clients who prefer to do invoices with negative amounts instead of creating refunds. The commits in this PR would give incorrect results for these grids.

### Grid 48/49
I saw no impact of the commites. That's strange: the invoices with negative amounts are in the results, while I would expect not to. Does this expression

`bal[][('tax_ids.tag_ids.id', '=', ref('l10n_be.tax_tag_49').id), '|', ('invoice_id.type', '=', 'out_refund'), ('credit', '=', 0)]`

means that the amount is taken into account only in case of refund?

### Grille 81/82/83/85
Tested with grid 82, based on "TVA Entrant - Frais de voiture - VAT 50% Non Deductible (Price Excl.)"
The result is correct : the base and the non deductible amount (50%) are taken into account in the grid 82, which was not the case before (only the base was taken into account).

### Grid 84, 86, 87, 88, 54, 55, 56, 57, 61, 63, 59, 64
No change in the code, not tested.

### Grid 62
I had no VAT with this tag in my test data. Not tested.
[Notice décla TVA.pdf](https://github.com/OCA/l10n-belgium/files/2327903/Notice.decla.TVA.pdf)

sbidoul:
@vdewulf About credit notes, my proposal was meant to cover invoices containing negative lines (such as a discount), but with a positive total. Without that, these lines end up declared in grid 49. 

To me an invoice with a negative total should be a credit note, no? But I'm just an engineer :) I'll check again with an accountant.

sbidoul:
@vdewulf here are some more fixes and improvements.

vdewulf:
Tested OK with the last commits.
Let's merge!

PR #66, opened 2018-07-11
[10.0][ADD] l10n_be_invoice_bba_supplier

rousseldenis:


rousseldenis:
@sbidoul @luc-demeyer

sbidoul:
@rousseldenis this is not BE specific. This must go to account-invoincing.

account_invoice_supplier_reference_type or something.

rousseldenis:
@sbidoul It is! https://github.com/odoo/odoo/tree/10.0/addons/l10n_be_invoice_bba

sbidoul:
Ok, my bad. I was convinced the reference_type field was visible without l10n_be_invoice_bba. This makes little sense to me to have the field in account and add it in the view in l10n_be. Anyway...

Can you please reopen your PR?

l10n_be_invoice_bba_supplier/readme/USAGE.rst
```diff
@@ -0,0 +1 @@
+Just install this module
```
sbidoul, line 1:
```
The reference type is visible just before the reference field on vendor bill forms. 
This module is best used with ``account_invoice_supplier_ref_unique`` which makes 
the distinction between supplier invoice reference (in a new field) and the payment reference.
```

luc-demeyer:
cf. https://github.com/luc-demeyer/noviat-apps/tree/10.0/l10n_be_invoice_bba_supplier

This module does the job, in is stable since installed already at multiple sites.
We could move this one to l10n_be_invoice_bba and think about a more generic solution for 11.0 (adding supplier payment ref & ref type fields in account-financial-tools and belgian add-on in l10n-belgium

PR #70, opened 2018-10-01
[9.0][IMP] l10n_be_mis_reports: auto expand account details

sbidoul:
We expand account detail by default in the template. It can be disabled
in the report instance, with recent mis builder versions.


sbidoul:
@vdewulf here is the 9.0 PR to add account details in P&L and Balance Sheet

sbidoul:
And now with the account details on the VAT declaration too.

vdewulf:
Tested OK.

PR #72, opened 2018-10-02
[TRANSLATE] Adding the English Text for each grid

vdewulf:
I added some english text on the grid description to have something more meaningful for the end user.


sbidoul:
@vdewulf looks good. May I suggest to remove the `Grid  ` and `Cadre ` prefixes which do not bring must added value?

vdewulf:
@sbidoul I would like to let these to match with the codes used in the official template of the VAT declaration. The accountant mentioned to me the references to this grid, that's why I let them.
Would you accept to let it like that?
Thanks for the review!

sbidoul:
The grid codes are fine and must stay of course, but I find the "Grid" word at the beginning of each line a bit repetitive. What do you think?

PR #86, opened 2020-03-08
[10.0][FIX]partner_kbo_bce

luc-demeyer:
base module, res_partner.py drops 'is_company' from vals hence we need to save vals before super()


sbidoul (approved):
Thanks, @luc-demeyer.

Do you know if a simple test would reproduce this bug? Would be useful to have it to detect changes across version.

luc-demeyer:
@sbidoul 
adding such a test would indeed be usefull, I'll put it on my todo list for the 13.0 version of this module.

PR #91, opened 2020-07-31
[13.0][MIG] l10n_be mis reports

sbidoul:


sbidoul:
This pure data module is identical across all versions, so I'll fast track this migration.

/ocabot merge patch

PR #105, opened 2021-04-06
[12.0][MIG] companyweb base

xavier-bouquiaux:
@sbidoul


sbidoul:
Looks good. Could you squash the migration commits ?

sbidoul (approved):

companyweb_base/models/res_partner.py
```diff
@@ -518,7 +518,7 @@ def cweb_button_enhance(self):
         )
         if r["StatusCode"] in [101, 302]:
             return self._cweb_call_wizard_credentials("Enter Companyweb credentials")
-        elif r["StatusCode"] != 0:
+        if r["StatusCode"] != 0:
```
rousseldenis: @xavier-bouquiaux Why changing elif here in the migration commit ?

rousseldenis (approved):

companyweb_base/models/res_partner.py
sbidoul: I changed this to keep the code aligned across versions

PR #107, opened 2021-04-07
[14.0][FIX] companyweb_base: don't import tests

sbidoul:


sbidoul:
/ocabot merge patch

cc/ @xavier-bouquiaux

PR #110, opened 2021-05-03
[13.0] [FIX] l10n_be_partner_identification: duplicated id codes

ThomasBinsfeld:
The code of partner identification category is [limited to 16 characters](https://github.com/OCA/partner-contact/blob/13.0/partner_identification/models/res_partner_id_category.py#L23). Resulting in duplicated codes for the two categories created by this module (they both begin with "L10n_be_id_numbe"), which made the methods `_compute_identification` and `_inverse_identification` crash.

In this PR, I fixed the two categories with shorter codes. It may break existing installations!

### DIFF
--- a/l10n_be_partner_identification/data/l10n_be_partner_identification.xml
+++ b/l10n_be_partner_identification/data/l10n_be_partner_identification.xml
@@ -5,7 +5,7 @@
         id="l10n_be_national_registry_number_category"
     >
         <field name="name">Belgium national registration number</field>
-        <field name="code">L10n_be_id_number</field>
+        <field name="code">be_nat_reg_num</field>
         <field
             name="validation_code"
         ><![CDATA[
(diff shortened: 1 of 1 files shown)


sbidoul:
Why not increasing the size limit ?

ThomasBinsfeld:
> Why not increasing the size limit ?

IMO it's the responsibility of this module to be compatible with the base module (partner_identification). I don't know what would be the side effects of changing the max size of the code in partner_identification. If there are other modules with >16 chars codes in updatable data files like here, their code will change at the update.

l10n_be_partner_identification/data/l10n_be_partner_identification.xml
```diff
@@ -5,7 +5,7 @@
         id="l10n_be_national_registry_number_category"
     >
         <field name="name">Belgium national registration number</field>
-        <field name="code">L10n_be_id_number</field>
+        <field name="code">be_id_number</field>
```
sbidoul: Ok, while you are at it:

```suggestion
        <field name="code">be_nat_reg_num</field>
```

sbidoul (approved):
cc/ @lmignon @rousseldenis

rousseldenis (approved):

lmignon (approved):

PR #113, opened 2021-07-19
[14.0][ADD] companyweb payment info xbo

xavier-bouquiaux:
@sbidoul


companyweb_payment_info/readme/CREDITS.rst
```diff
@@ -0,0 +1,9 @@
+This module has been originally developed by `ACSONE <https://acsone.eu>`_ for
+Odoo version 8.
+
+It was subsequently enhanced by Jacques-Etienne Baudoux (BCIM sprl). The data
+model of that version has been preserved to facilitate porting.
+
+In 2021, the module was redeveloped under the current form, to support the new
+Companyweb SOAP api. This development has been funded by `Companyweb
+<https://www.companyweb.be>`__.
```
sbidoul, line 3: Don't forget to update the docs.

companyweb_payment_info/wizards/payment_info_wizard.py
```diff
@@ -0,0 +1,345 @@
...
+import re
+
+import zeep
+
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError
+
+import odoo.addons.companyweb_base.models.res_partner as cweb_partner
+
+CWEB_INVOICE_MAX_LEN = 2
```
sbidoul: TODO

companyweb_payment_info/wizards/payment_info_wizard.py
```diff
@@ -0,0 +1,345 @@
...
+    def _check_group(self):
+        if not self.env.user.has_group("companyweb_payment_info.cweb_upload"):
+            raise UserError(
+                _(
+                    """Companyweb : You need to be in the security group :
+                    Upload Companyweb Invoices to perform this action"""
+                )
+            )
+        else:
+            return True
```
sbidoul:
```suggestion
```

No need to return True here since the caller does not check the return value.

companyweb_payment_info/wizards/payment_info_wizard.py
```diff
@@ -0,0 +1,345 @@
...
+
+    def _create_step1_summary(self, invoice_to_send, last_period_sent, period_to_send):
+        summary = _(
+            "<h2>CompanyWeb upload</h2>"
+            "You are about to submit <strong>%d</strong> open invoices <br/>"
+            "to Companyweb for the company <strong>%s</strong>.<br/>"
+            "The previous period that was sent is <strong>%s</strong>.<br/>"
+            "Make sure you have closed your bank statements for period <strong>%s</strong>.",
+            len(invoice_to_send),
+            self.env.user.company_id.name,
```
sbidoul: escape html

companyweb_payment_info/wizards/payment_info_wizard.py
```diff
@@ -0,0 +1,345 @@
...
+            last_period_sent,
+            str(period_to_send.year) + str(period_to_send.month),
+        )
```
sbidoul, line 224: This should be `_("... message ...") % (var1, var2, ...)`

companyweb_payment_info/wizards/payment_info_wizard.py
```diff
@@ -0,0 +1,345 @@
...
+        if response_previous_period["StatusCode"] in [11, 15, 16]:
+            return self._cweb_call_wizard_credentials(_("Bad Credentials"))
+        last_date_sent = "%d%02d" % (
+            response_previous_period["PreviousYear"],
+            response_previous_period["PreviousMonth"],
+        )
+        period_to_send = self._get_period_to_send()
+        invoices_to_send = self._create_invoices_to_send(client)
+        if len(invoices_to_send) == 0:
+            raise UserError(_("Companyweb : No Invoices to send"))
```
sbidoul, line 94: `_create_invoices_to_send` already raises an exception in some cases when there are no open invoices. So for consistency that check should either be done only in _create_invoices_to_send or only here in the caller.

companyweb_payment_info/wizards/payment_info_wizard.py
```diff
@@ -0,0 +1,345 @@
...
+        invoice_to_send = []
+
+        for invoice in open_invoices:
+            if not invoice.partner_id.vat:
+                continue
+            if not re.match(
+                r"^(BE)?\s?0?[0-9]{4}\s?\.?[0-9]{3}\s?\.?[0-9]{3}$",
+                invoice.partner_id.vat,
+            ):
+                continue
```
sbidoul, line 275: For performance, the regex should be compiled once at module load time, so:

```python
# in module
VAT_RE = re.compile(r"^(BE)?\s?0?[0-9]{4}\s?\.?[0-9]{3}\s?\.?[0-9]{3}$")

for ...:
  if not VAT_RE.match(invoice.partner_id.vat):
      continue
```

companyweb_payment_info/wizards/payment_info_wizard.py
```diff
@@ -0,0 +1,345 @@
...
+    def _cweb_send_batch(self, client, invoices_to_send, transaction_key):
+
+        invoice_list_type = client.get_type("ns0:ArrayOfInvoiceRequest")
+
+        it = self._cweb_split_invoice_list(invoices_to_send)
+        while True:
+            try:
+                invoices_request_to_send_splitted = next(it)
+            except StopIteration:
+                break
```
sbidoul:
```suggestion
        for invoices_request_to_send_splitted in self._cweb_split_invoice_list(invoices_to_send):
```

companyweb_payment_info/wizards/payment_info_wizard.py
```diff
@@ -0,0 +1,345 @@
...
+        return response_previous_period
+
+    def _get_module_version(self):
+        return (
+            self.env["ir.module.module"]
+            .search([("name", "=", "companyweb_payment_info")])
+            .installed_version
+        )
+
+    def _cweb_split_invoice_list(self, full_invoice_list):
```
sbidoul: Nitpick: this is a utility function that is not specific to invoice, so this would be better declared at module level and with a generic name, such as `def _chunks(self, lst, chunk_size)`.

companyweb_base/models/res_partner.py
```diff
@@ -517,7 +520,9 @@ def cweb_button_enhance(self):
             )
         )
         if r["StatusCode"] in [101, 302]:
-            return self._cweb_call_wizard_credentials("Enter Companyweb credentials")
+            return self._cweb_call_wizard_credentials(
+                "Enter Companyweb credentials", self.env.context
+            )
         elif r["StatusCode"] != 0:
             raise UserError(
                 _("Companyweb status : %s : %s ", r["StatusCode"], r["StatusMessage"])
```
sbidoul:
```suggestion
                _("Companyweb status : %s : %s ") % (r["StatusCode"], r["StatusMessage"])
```

companyweb_payment_info/wizards/payment_info_wizard.py
```diff
@@ -0,0 +1,345 @@
...
+
+        self._check_group()
+
+        supplierVat = self.env.user.company_id.vat
+        if not supplierVat:
+            raise UserError(
+                _(
+                    "Companyweb : You need to set the tva field for the current company : %s",
+                    self.env.user.company_id.name,
+                )
```
sbidoul:
```suggestion
                _("Companyweb : You need to set the tva field for the current company : %s") %
                (self.env.user.company_id.name, )
```

companyweb_payment_info/wizards/payment_info_wizard.py
```diff
@@ -0,0 +1,345 @@
...
+        client = zeep.Client(
+            "https://payex.companyweb.be/v1/PaymentExperienceService.asmx"
+        )
+
+        invoices_to_send = self._create_invoices_to_send(client)
+        result_start_transaction = self._cweb_start_transaction(client)
+        if result_start_transaction["StatusCode"] != 0:
+            raise UserError(
+                _(
+                    "Error from Companyweb : %s : %s",
```
sbidoul: here too

companyweb_payment_info/wizards/payment_info_wizard.py
```diff
@@ -0,0 +1,345 @@
...
+                    response_summary.InvoicesSummary["LinesWithInvalidExpirationDate"],
+                    response_summary.InvoicesSummary["LinesWithInvalidInvoiceDate"],
+                    response_summary.InvoicesSummary["LinesWithInvalidInvoiceNumber"],
+                    response_summary.InvoicesSummary["LinesWithInvalidOpenAmount"],
+                    response_summary.InvoicesSummary["LinesWithInvalidVatNumber"],
+                )
+            )
+        return response_summary
+
+    def _create_step2_resume(self, invoice_summary):
```
sbidoul:
```suggestion
    def _create_step2_summary(self, invoice_summary):
```

companyweb_payment_info/wizards/payment_info_wizard.py
```diff
@@ -0,0 +1,345 @@
...
+        summary = _(
+            "<h2>CompanyWeb upload Status</h2>"
+            "Here under the status of your Companyweb's upload <br/>"
+            "You sent <strong>%d</strong> invoices.<br/>",
```
sbidoul: %

sbidoul:
@xavier-bouquiaux can you check the travis error ? It looks like the companyweb_base test is trying to send the request to companyweb, so the vcr cassete might be broken ?

companyweb_base/models/res_partner.py
```diff
@@ -547,15 +551,16 @@ def cweb_button_copy_address(self):
             message=_("Companyweb Address Successfully copied")
         )
 
-    def _cweb_call_wizard_credentials(self, wizard_name):
+    @api.model
+    def _cweb_call_wizard_credentials(self, wizard_name, context):
```
sbidoul: accepting the context as argument sounds weird ? Why not use `self.env.context` ?

companyweb_base/models/res_partner.py
```diff
@@ -547,15 +551,16 @@ def cweb_button_copy_address(self):
...
             "name": wizard_name,
             "type": "ir.actions.act_window",
             "view_type": "form",
             "view_mode": "form",
-            "res_model": "companyweb_base.credential_wizard",
+            "res_model": "companyweb_base.credential_wizard_base",
             "view_id": wizard_form.id,
             "target": "new",
-            "context": self.env.context,
+            "context": context,
```
sbidoul: Accepting the context as method argument sounds weird. Why not use `self.env.context` ?

companyweb_payment_info/wizards/payment_info_wizard.py
```diff
@@ -0,0 +1,393 @@
...
+
+    def payment_info_entry_point(self):
+        if self.wizard_step == "step1":
+            return self._cweb_payment_info_step1()
+        elif self.wizard_step == "step2":
+            return self._cweb_payment_info_step2()
+        elif self.wizard_step == "step3":
+            return True
+        else:
+            raise UserError(_("Unexpected step %s") % self.wizard_step)
```
sbidoul: If this happens it's not the user's fault, it's a bug, so I'd use SystemError here.

companyweb_payment_info/wizards/payment_info_wizard.py
```diff
@@ -0,0 +1,393 @@
...
+    @api.model
+    def _cweb_payment_info_step1(self):
+
+        self._check_group()
+
+        supplierVat = self.env.user.company_id.vat
+        if not supplierVat:
+            raise UserError(
+                _(
+                    "Companyweb : You need to set the tva field for the current company : %s"
```
sbidoul:
```suggestion
                    "Companyweb : You need to set the vat field for the current company : %s"
```

companyweb_payment_info/wizards/payment_info_wizard.py
```diff
@@ -0,0 +1,393 @@
...
+        invoice_to_send = []
+
+        for invoice in open_invoices:
+            if not invoice.partner_id.vat:
+                continue
+            if not VAT_RE.match(invoice.partner_id.vat):
+                continue
+
+            to_send = invoice_request_type(
+                InvoiceNumber=invoice.id,
```
sbidoul, line 278: Is it not better to send the real invoice number here (the account move number?)

companyweb_payment_info/wizards/payment_info_wizard.py
```diff
@@ -0,0 +1,393 @@
...
+                VatNumber=invoice.partner_id.vat,
+            )
+            invoice_to_send.append(to_send)
+        return invoice_to_send
+
+    def _get_open_invoices(self):
+        return self.env["account.move"].search(
+            [
+                ("payment_state", "=", "not_paid"),
+                ("state", "=", "posted"),
```
sbidoul, line 291: Don't we need to filter on customer invoices only ?
What about customer refunds ?

sbidoul (commented):
A few comments.

Also the documentations needs to be written (the `readme` directory still contains the docs of the base module).

companyweb_payment_info/readme/USAGE.rst
```diff
@@ -0,0 +1,11 @@
...
+
+* Upload Companyweb Data : can upload invoice to Companyweb
+
+In order to use this module you should
+* be in the right security group
+* set a vat for the company you are logged in
+* have companyweb credentials
+* have open invoices for customer with a belgian's vat
+
+The module's wizard helps you to configure what's needed
```
sbidoul, line 11: Perhaps say a word about where the wizard can be found in the menu ?

sbidoul: About the docs: it seem the images added in `companyweb_payment_info/static/description` come from `companyweb_base` and are not related to this new module ?

companyweb_payment_info/wizards/payment_info_wizard.py
```diff
@@ -0,0 +1,396 @@
...
+        self._check_group()
+
+        supplierVat = self.env.user.company_id.vat
+        if not supplierVat:
+            raise UserError(
+                _(
+                    "Companyweb : You need to set the vat field for the current company : %s"
+                    % self.env.user.company_id.name,
+                )
+            )
```
sbidoul, line 64: Should we also check that the VAT number starts with BE ?

companyweb_payment_info/wizards/payment_info_wizard.py
```diff
@@ -0,0 +1,396 @@
...
+        return invoice_to_send
+
+    def _get_open_invoices(self):
+        return self.env["account.move"].search(
+            [
+                ("payment_state", "=", "not_paid"),
+                ("state", "=", "posted"),
+                ("move_type", "in", ["out_invoice"]),
+            ]
+        )
```
sbidoul, line 295: Don't we need to filter on the company here ?

companyweb_payment_info/readme/USAGE.rst
xavier-bouquiaux: Done

companyweb_payment_info/wizards/payment_info_wizard.py
```diff
@@ -0,0 +1,396 @@
...
+                ),
+                TransactionKey=transaction_key,
+            )
+        )
+        return response_summary
+
+    def _create_step2_summary(self, invoice_summary):
+        summary = _(
+            "<h2>Companyweb upload Status</h2>"
+            "Here under a small resume, the full resume "
```
sbidoul: summary instead of resume

companyweb_payment_info/wizards/payment_info_wizard.py
xavier-bouquiaux: Filter only on out_invoice done

companyweb_payment_info/wizards/payment_info_wizard.py
```diff
@@ -0,0 +1,396 @@
...
+            "will be send to <strong>%s</strong> <br/>"
```
sbidoul: s/send/sent/

companyweb_payment_info/wizards/payment_info_wizard.py
sbidoul: short summary instead of small resume, actually

xavier-bouquiaux: done & i18n also

companyweb_payment_info/wizards/payment_info_wizard.py
```diff
@@ -0,0 +1,396 @@
...
+        )
+        if response_previous_period["PreviousPeriodExists"]:
+
+            summary = _(
+                "<h2>Companyweb upload</h2>"
+                "You are about to submit <strong>%s</strong> open invoices <br/>"
+                "to Companyweb for the company <strong>%s</strong>.<br/>"
+                "The previous period that was sent is <strong>%s</strong>.<br/>"
+                "Your odoo login : <strong>%s</strong> will receive a full "
+                "resume at the end of this transaction <br/>"
```
sbidoul: summary

companyweb_payment_info/wizards/payment_info_wizard.py
```diff
@@ -0,0 +1,396 @@
...
+                )
+            )
+        else:
+            summary = _(
+                "<h2>Companyweb upload</h2>"
+                "You are about to submit <strong>%s</strong> open invoices <br/>"
+                "to Companyweb for the company <strong>%s</strong>.<br/>"
+                "<strong>It's the first time you use this feature "
+                "for this company</strong><br/>"
+                "Your odoo login : <strong>%s</strong> will received a full resume "
```
sbidoul: summary

companyweb_payment_info/wizards/payment_info_wizard.py
xavier-bouquiaux: done

companyweb_payment_info/wizards/payment_info_wizard.py
xavier-bouquiaux: done

PR #116, opened 2021-09-23
[14.0][MIG] account_statement_import_Coda

KristofGommers:
Hello,

I'm new to odoo/github so I hope i'm doing this right.

Here is my pull request for importing coda.(Built on V13).

I fixed a few fields , and I am able to import coda's now in V14 if you have account_statement_import.

Sincerly
Kristof


qgroulard:
Could you please follow OCA Migration guidelines ?
https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-14.0

A cleaner commits history would ease review.
Thanks

sbidoul (changes requested):
@KristofGommers since there is a module named `account_bank_statement_import_coda`, the best is to migrate it, following the procedure explained in #93 

Here it seems you are creating a module with a different name.

marielejeune:
@sbidoul module was named account_bank_statement_import_coda in v13 since it depended from account_bank_statement_import that was in Odoo Community. But this module is now part of Odoo Enterprise and we now have to depend from OCA/bank-statement-import/account_statement_import since v14. The name of the OCA module was changed in order to avoid a conflict with the Enterprise module, hence it would be logical to do the same in this sub-module.

sbidoul:
Ah, yes, thanks for the explanation @marielejeune. 

So the best approach is probably to migrate account_bank_statement_import_coda using the regular migration procedure (preserving the commit history), and then doing the renaming in the migration commit.

PR #128, opened 2022-01-20
[12.0] fix partner vat listing file extension

robinkeunen:
**Before**: no extension specified in the filename, svg is added by default
**After**: append .xml to `name` used as file name


carmenbianca (approved):
LGTM! I was going to try this fix as well, but I couldn't get this module to run locally for some reason.

houssine78 (approved):

PR #130, opened 2022-02-01
[14.0] [MIG] account_statement_import_coda

marielejeune:


marielejeune:
Migration of account_bank_statement_import_coda from v13 to v14. Changed the name of the module to follow what was done for account_statement_import.

sbidoul (approved):

houssine78 (approved):
LGTM

houssine78:
@robinkeunen I think we have a PR ready to me merged here

gva-acsone (approved):
LGTM

account_statement_import_coda/wizard/account_statement_import_coda.py
```diff
@@ -0,0 +1,218 @@
...
+                    'date':value,
+                    'amount':value,
+                    'payment_ref':value,
+                }
+        """
+        amount = line.transaction_amount
+        if line.transaction_amount_sign == AmountSign.DEBIT:
+            amount = -amount
+        return {
+            "name": self.get_st_line_name(line, globalisation_dict),
```
alexis-via, line 198: Now that account.bank.statement.line inherit from account.move, you're not supposed to force the "name" in this dict! Otherwise, the move will not have the "name" set by this code instead of having the "name" set by the journal like all the other moves. You should just set "payment_ref".

marielejeune: Hi @alexis-via. I agree with you. A PR was opened several months ago with this fix, but wasn't approved yet. See https://github.com/OCA/l10n-belgium/pull/139

Btw, a second PR with another fix is here, could you also have a look please? https://github.com/OCA/l10n-belgium/pull/140

PR #138, opened 2022-04-02
[FIX] l10n_be_vat_reports: round amounts and format to 2 decimals

robinkeunen:
Belgian administration only accepts 2 and only 2 decimals in the reports.
Amounts are already rounded in the query to make sure total sums match.


remytms (approved):
LGTM.

PR #150, opened 2022-10-05
[12.0][ADD] l10n_be_cooperator

huguesdk:
move `l10n_be_cooperator` module from [coopiteasy/vertical-cooperative](https://github.com/coopiteasy/vertical-cooperative) to the oca.

not sure whether this should go in this repository or in [OCA/cooperative](https://github.com/OCA/cooperative), though. first created OCA/cooperative#10, then now this here also in case this is the right place.


robinkeunen:
I think it has its place here since it is related to Belgian accounting and Belgian legal certificates (tax shelter, ...). Let's see what @luc-demeyer or @sbidoul think about it.

robinkeunen:
@huguesdk maybe also ask in l10n-france or l10n-switzerland so its consistent accross localizations.

l10n_be_cooperator/tests/test_tax_shelter.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def _create_dummy_cooperator_2021(self):
+        vals = self.get_dummy_subscription_requests_vals()
+        vals["date"] = date(2021, 6, 21)
+        subscription_request = self.env["subscription.request"].create(vals)
+        subscription_request.validate_subscription_request()
+        self.pay_invoice(
+            subscription_request.capital_release_request,
+            payment_date=date(2021, 6, 22),
+        )
```
robinkeunen, line 24: thx for all these helpers 👍

robinkeunen (approved):
👍 but didn't you want to move `send_mail_with_additional_attachments` to another module ?

huguesdk:
> +1 but didn't you want to move `send_mail_with_additional_attachments` to another module ?

that would be good, indeed, but the current implementation would not work if `force_send` is True. i wonder whether there would be a better way to add multiple attachments to a mail template. or are we doing it wrong?

robinkeunen:
@huguesdk As discussed for `send_mail_with_additional_attachments`, I suggest :

1. leaving a note to explain why its _hacky_ and improvement suggestions
2. merge this as is
3. if better ways of doing this in 14+, leave as is in 12 otherwise let's suggest a module to the oca

PR #151, opened 2022-10-14
[14.0]add l10n_be_intrastat_product

luc-demeyer:


robinkeunen:
Hi Luc, thank you for this. I don't have enough knowledge on these reports to review the code. How can I help ?

luc-demeyer:
@robinkeunen 
These reports are not easy and not all use cases are implemented since a completely correct implementation would requires linking all invoicing data to stock moves and hence also associated processes at customer site.
I try to make here a report which corresponds to current practices on the belgian market and exceptions can always be corrected via the 'declaration lines' before submitting the report to the authorities.

One of our customers is currently testing this latest version of the declaration.
As soon as I get positive feedback I'll merge the code.

nayatec (approved):

sbidoul:
Again here, please never merge manually. Use the ocabot merge command.

PR #155, opened 2022-11-11
intrastat readme

luc-demeyer:

PR #156, opened 2022-11-11
use README.rst from maintainer tools templates

luc-demeyer:


sbidoul:
Please never merge manually. Use the ocabot merge command.
In this case, updating the README manually is useless as it will be re-generated by the bot.

luc-demeyer:
@sbidoul 
I actually do know that I shouldn't merge manually. 
In this case it was just the frustration of the failing https://github.com/OCA/l10n-belgium/pull/155 with an error message that's not very helpfull. I didn't do the effort to read the oca-gen-addon-readme code in order to figure out if this code replaces or updates the exsiting README.rst.

I think the problem is that the local pre-commit doesn't generate the README.rst hence we need to wait upon the result of the ocabot merge to detect this kind of problems. Would it be difficult to change the 'dotfiles' so that we README.rst is generated locally and hence we can fix such errors before creating the PR ?

sbidoul:
Yeah I know it can be frustrating. However when a red build is merged, this frustrates all other contributors :)

I think we considered generating the README.rst in pre-commit but rejected the idea. I don't remember why right now.
What we could do however is check the generated README syntax without generating it. We could also try to suggest the fix as the problem you faced is quite common.

sbidoul:
Created https://github.com/OCA/oca-addons-repo-template/issues/169 to track the idea.

luc-demeyer:
@sbidoul 
>>> Yeah I know it can be frustrating. However when a red build is merged, this frustrates all other contributors :)
I merged a green build (this one only became red when using the ocabot merge) but I fully agree that nobody should bypass the ocabot merge. The good nesw is that the repo has been fixed before someone else got impacted (but that's because there is little activity on this repo).

Thanks for considering changing the toolset to minimise README.rst generation isssue for other contributors (I have spent enough time on this one hence I will not hit it again).

PR #160, opened 2022-12-02
[MIG][16.0] companyweb

xavier-bouquiaux:
migration of companyweb's module to odoo 16


companyweb_base/views/partner_view.xml
```diff
@@ -74,7 +74,6 @@
                             string="Global"
                             attrs="{'invisible':[('cweb_name_enable', '!=', True )]}"
                         >
-                            <div /> <!-- here it is -->
```
sbidoul: @xavier-bouquiaux out of curiosity, what was this div for in previous version?

PR #163, opened 2022-12-06
[14.0][FIX] update repository template to 1.11.1

huguesdk:


.copier-answers.yml
```diff
@@ -14,8 +14,8 @@ odoo_version: 14.0
 org_name: Odoo Community Association (OCA)
 org_slug: OCA
 rebel_module_groups: []
-repo_description: 'TODO: add repo description.'
-repo_name: l10n-belgium
+repo_description: These modules complement the official l10n_be.
+repo_name: Odoo modules for Belgium
```
robinkeunen: Maybe rather _Belgian localization_ (like in l10n-france), i find it more precise

robinkeunen (changes requested):

.copier-answers.yml
huguesdk: i just copied the description from the 13.0 branch. your proposition is better indeed. fixed!

robinkeunen (approved):

PR #177, opened 2023-01-04
12.0 [ADD] l10n_be_national_number

victor-champonnois:
Minimal implementation of the Belgian National Number in partner_identification.
The main Roadmap is to add a validation method.


l10n_be_national_number/models/__init__.py
```diff
@@ -0,0 +1,2 @@
+from . import res_partner
+from . import res_partner_id_category
```
robinkeunen: I do not see this file in the PR. Did you forget to include it ?

robinkeunen (changes requested):

l10n_be_national_number/models/__init__.py
victor-champonnois: fixed

victor-champonnois:
@robinkeunen fixed and rebase

robinkeunen (approved):
LGTM

PR #180, opened 2023-02-14
[FIX] l10n_be_cooperator: tax shelter cron

robinkeunen:
The module name was missing in the model_id field. Odoo thus
failed to find the model upon cron execution.

Since the data is set as nocreate=True, we need a pre-migration
script to force data recreation.


carmenbianca (approved):
LGTM

PR #200, opened 2023-09-04
[FIX] companyweb_payment_info: Correct translations

rousseldenis:
@OCA/local-belgium-maintainers


sbidoul:
Thanks Looks trivial, so fast tracking.

/ocabot merge patch

sbidoul:
Hm. Weird error. It is merged correctly though. Probably a side effect of the new README generator creating a commit that updates the digest in README.rst because the manifest was updated with the new version number. Something to chew on for me...

PR #203, opened 2023-09-15
[14.0][ADD] MIS Reports templates for Balance Sheet and Profit & Loss

mathisjacoby:
New reports in `l10n_be_mis_reports` for the micro, abridged and full model templates for the Balance Sheet and the Profit & Loss for :
- Companies with capital
- Companies without capital
- Non-profit institution

The templates are generated with two scripts located in the 'script' folder, which use JSON files containing data from the Central Balance Sheet Office taxonomy. 

They are available in English, French, Dutch and German.

The KPI names are structured for an other module that will come later. This module will allow to generate XBRL instances, ready to be submitted to the National Bank of Belgium.


l10n_be_mis_reports/__manifest__.py
```diff
@@ -13,9 +13,32 @@
...
+        "data/mis_report_bs_m87-f.xml",
+        "data/mis_report_pl_m87-f.xml",
+        "data/mis_report_bs_deprecated.xml",
+        "data/mis_report_pl_deprecated.xml",
         "data/mis_report_vat.xml",
     ],
     "installable": True,
+    "external_dependencies": {
+        "python": [
+            "click",
```
sbidoul: I'd remove this, since it is only necessary for generating the template. So let's just document in the README of the script directory.

sbidoul:
@mathisjacoby could you do it for 16.0 too?

mathisjacoby:
@sbidoul Of course. Here is the PR for 16.0 : https://github.com/OCA/l10n-belgium/pull/205

l10n_be_mis_reports/script/README.rst
```diff
@@ -0,0 +1,45 @@
+correct_xml and generate_xml are two scripts that allow to generate a MIS Report Templates for the Balance Sheet and the Profit & Loss
+
+First, use the generate_xml script to generate the template of the model that you need. 
+Then, use the correct_xml script which will make some modification to the template previously generated.
```
sbidoul: It would be nice to have a script that regenerates everything.

PR #204, opened 2023-09-21
[16.0][MIG] l10n_be_iso20022_pain

sbejaoui:
needs: https://github.com/OCA/bank-payment/pull/1314


l10n_be_iso20022_pain/models/account_move.py
```diff
@@ -25,7 +25,8 @@ class AccountMove(models.Model):
     _inherit = "account.move"
 
     reference_type = fields.Selection(
-        selection_add=[("bba", "BBA Structured Communication")]
+        selection_add=[("bba", "BBA Structured Communication")],
```
rousseldenis: @sbejaoui Is this still needed following this:

https://github.com/odoo/odoo/blob/ec5f6003f8f6f8de16c6c8b2b023a0511a55d498/addons/account/models/account_move.py#L2605

And this:

https://github.com/odoo/odoo/blob/ec5f6003f8f6f8de16c6c8b2b023a0511a55d498/addons/l10n_be/models/account_move.py#L20C9-L20C42

sbejaoui: Yes, in the previous versions, `_get_invoice_reference_be_partner` was located in `l10n_be_invoice_bba`, but now it has been merged into `l10n_be`. This method provides a default BBA-formatted reference.

`l10n_be_iso20022_pain` add a new reference type `bba` to ensure that the reference (whether generated or edited by the user) complies with the valid BBA format.

rousseldenis (approved):

l10n_be_iso20022_pain/models/account_move.py
sbidoul: I don't think we should add a reference_type. If `reference_type` is `structured` and `journal_id.invoice_reference_model` is `be` it means we have a bba structured communication.

l10n_be_iso20022_pain/models/account_move.py
```diff
@@ -0,0 +1,70 @@
...
+        if mod == int(bbacomm[-2:]):
+            return True
+    return False
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    reference_type = fields.Selection(
+        selection_add=[("bba", "BBA Structured Communication")],
```
luc-demeyer: "bba" in payment order should be "BBA" (uppercase)

ThomasBinsfeld (approved):

ThomasBinsfeld:
LGTM

ThomasBinsfeld (approved):

l10n_be_iso20022_pain/models/account_move.py
sbejaoui: Correct, 

I believe we should keep only the check for the ref in this module.

cc @ThomasBinsfeld

ThomasBinsfeld (approved):

sbidoul:
@sbejaoui @ThomasBinsfeld Is it not necessary to check and/or enforce `invoice_reference_model = 'be'`, for instance in the migration script or in `_check_communication` ?

l10n_be_iso20022_pain/models/account_payment_line.py
```diff
@@ -0,0 +1,19 @@
...
+class AccountPaymentLine(models.Model):
+    _inherit = "account.payment.line"
+
+    @api.constrains("communication", "communication_type")
+    def _check_communication(self):
+        for rec in self:
+            if rec.communication_type == "structured" and not check_bbacomm(
+                rec.communication
+            ):
+                raise ValidationError(_("Invalid BBA Structured Communication !"))
```
ThomasBinsfeld:
```suggestion
    @api.constrains("order_id", "communication", "communication_type")
    def _check_communication(self):
        for rec in self:
            if (
                rec.order_id.journal_id.invoice_reference_model == "be"
                and rec.communication_type == "structured"
                and not check_bbacomm(rec.communication
            ):
                raise ValidationError(_("Invalid BBA Structured Communication !"))
```

ThomasBinsfeld:
> @sbejaoui @ThomasBinsfeld Is it not necessary to check and/or enforce `invoice_reference_model = 'be'`, for instance in the migration script or in `_check_communication` ?

The migration script looks to me. But the constraint on `account.payment.line` should check the reference model of the journal indeed.

ThomasBinsfeld (changes requested):

l10n_be_iso20022_pain/models/account_payment_line.py
sbejaoui: @ThomasBinsfeld , I made the field invoice_reference_model visible for the bank journal so we can perform this check. See the last commit.

ThomasBinsfeld (approved):

l10n_be_iso20022_pain/views/account_journal.xml
```diff
@@ -0,0 +1,23 @@
...
+        <field name="model">account.journal</field>
+        <field name="inherit_id" ref="account.view_account_journal_form" />
+        <field name="arch" type="xml">
+            <xpath
+                expr="//field[@name='invoice_reference_type']/parent::group"
+                position="attributes"
+            >
+                <attribute
+                    name="attrs"
+                >{'invisible': [('type', 'not in', ('sale', 'bank'))]}</attribute>
```
sbidoul, line 16: I'm not sure it makes sense to make invoice_reference_model and invoice_reference_type visible for other journals than customer invoice journals. Because when we pay invoice bills with a journal, we don't control the type of reference we need to use for the suppliers. So I think invoice_reference_model and invoice_reference_type should only be used when generating payment reference on documents that we generate, not documents that we receive from suppliers.

l10n_be_iso20022_pain/models/account_move.py
```diff
@@ -0,0 +1,40 @@
...
+
+    @api.constrains("reference_type", "ref")
+    def _check_communication(self):
+        for rec in self:
+            if (
+                rec.journal_id.invoice_reference_model == "be"
+                and rec.journal_id.invoice_reference_type == "invoice"
+                and rec.reference_type == "structured"
+            ):
+                return check_bbacomm(rec.ref)
```
sbidoul, line 34: Should we not use `rec.payment_reference` here?

sbidoul: At least for vendor bills, where `ref` is the supplier invoice number?

l10n_be_iso20022_pain/models/account_move.py
sbidoul: @ThomasBinsfeld can you elaborate why you downvoted the comment above?
@luc-demeyer if you have a link to the spec of what must go in a SEPA Credit Transfer file for BBA structured communications, that would be helpful?

ThomasBinsfeld: I didn't see the need to change that and imply some refactoring. But this part has been removed since.

l10n_be_iso20022_pain/views/account_journal.xml
sbejaoui: In this case, we remove the check on the payment reference in payment orders. We assume that the reference was checked for our invoices, and we don't concern ourselves with whether it's correct for vendor bills.

l10n_be_iso20022_pain/models/account_move.py
sbejaoui: > Should we not use `rec.payment_reference` here?

yes

sbejaoui: > At least for vendor bills, where `ref` is the supplier invoice number?

only for out_invoice, and don't check for VB

remytms: @sbejaoui If you agree with @sbidoul on the `rec.payment_reference` can you make the changes ? :)

remytms:
If no reaction, the PR will be marked as stale.

PR #205, opened 2023-09-22
[16.0][ADD] MIS Reports templates for Balance Sheet and Profit & Loss

mathisjacoby:
New reports in `l10n_be_mis_reports` for the micro, abridged and full model templates for the Balance Sheet and the Profit & Loss for :
- Companies with capital
- Companies without capital
- Non-profit institution

The templates are generated with two scripts located in the 'script' folder, which use JSON files containing data from the Central Balance Sheet Office taxonomy. 

They are available in English, French, Dutch and German.

The KPI names are structured for an other module that will come later. This module will allow to generate XBRL instances, ready to be submitted to the National Bank of Belgium.


sbidoul:
I tested this on real data for companies with capital (abridged model).

I think this is good to go. Thanks @mathisjacoby!

@mathisjacoby could you just prefix the report template titles with `Belgium` and add translations for these titles?

mathisjacoby:
@sbidoul Thanks for the review ! 

It's done. May you check if it's correct now ?

sbidoul (approved):

vdewulf:
Hello,
Thanks for this work!
I compared the report for the balance sheet (full model) for the association with the model we created in our accounting company on version 12 community (manually) with the French language. I exported the reports in excel from both database. See attachment for full details.

I saw 2 differences
- in v16 report, 2 lines are missing (Actions propres [50] et Autres placements [51/53]), but the total line 50/53 is there. This is thus not an issue and we might have added it in v12 even though it was not necessary (as we can see in the PDF of the official declaration).

- the line "FONDS SOCIAL [10/15]" in our custom report is called in v16 "CAPITAUX PROPRES [10/15]", which is strange for an association who has no "capital". This is more of a (small) issue. The official PDF report available on the BNB website (https://www.nbb.be/doc/ba/models/npi/release%202023/release2021_fr_modele_complet_asbl_v8.pdf) indicates "Fonds social" as well.

I then created the report on the runboat database, and I see that active is different from passive section. This is, I suppose, linked to strange demo data on the accounting modules on runboat. I didn't investigate this.
[v16_OCA-module_bilan_association_full-model_comp_bilan_v12_Chaf.xlsx](https://github.com/OCA/l10n-belgium/files/14446789/v16_OCA-module_bilan_association_full-model_comp_bilan_v12_Chaf.xlsx)

Conclusion: I'm wondering if the translation is really up to date, but except that, I think we can proceed with the merge of this PR (translations are easy to fix afterwards with Weblate or in production database directly if needed).

sbidoul:
Thanks for the test @vdewulf. 

I suspect the translation issue may reveal something in the way transalations are generated. @mathisjacoby could you have a quick look?

mathisjacoby:
Hi, 

Thanks a lot for your review @vdewulf !

@sbidoul Of course, I'll have a look.

sbidoul:
@mathisjacoby have you had a chance to look into the translation issue?

mathisjacoby:
@sbidoul Not yet, sorry. I just need to finish something else before. 

Could it be next week or do you want me to check it now ?

sbidoul:
@vdewulf  the translation of Equity is a very special case.

First we have a limitation in Odoo where the export of the .pot file merges all terms of a module data files that have the same source language in English. So even if we split it manually to allow for a different translation of Equity for some reports, the CI will merge it again.

Second it seems that the translation of Equity for companies without capital (to Fonds Social) is only different in French (for Dutch and German we have the same translation as in other reports).

So there are two solutions: either we tweak the English term to be slightly different, or we just don't worry about it.

I'm going to merge as is, and we can address that in a followup.

/ocabot merge major

PR #211, opened 2023-11-30
[17.0][MIG] companyweb_base

mathisjacoby:
Migration of module `companyweb_base` based on 16.0

- Move keys hardcoded from code to an `ir_config_parameter` file
- Allow requests to companyweb API from tests


companyweb_base/tests/test_api_cweb.py
```diff
@@ -41,7 +54,7 @@ def _get_vcr_kwargs(self, **kwargs):
 
     @freeze_time("2021-03-23")  # because the login hash includes the date
     def test_cweb_button(self):
-        self.p1.with_context(lang="fr_Fr").cweb_button_enhance()
+        self.p1.with_context(lang="fr_FR").cweb_button_enhance()
```
sbidoul, line 64: The French translators community wants to remove `fr_FR` and use `fr` only..
@xavier-bouquiaux do you remember why we use fr_FR here?

companyweb_base/data/ir_config_parameter.xml
```diff
@@ -0,0 +1,18 @@
...
+    <!-- Companyweb is ok with those key being visible on github -->
+    <record id="service_integrator_id" model="ir.config_parameter">
+        <field name="key">service.integrator.id</field>
+        <field name="value">acsone</field>
+    </record>
+    <record id="service_integrator_secret" model="ir.config_parameter">
+        <field name="key">service.integrator.secret</field>
+        <field name="value">ECAB8ACF-9AE1-4E90-BD0D-05A1F47A3FE9</field>
+    </record>
+</odoo>
```
sbidoul, line 9: The service integrator id and secret is not meant to be changed by users or administrator. So I think we should keep it in the code.

companyweb_base/__manifest__.py
```diff
@@ -23,10 +24,9 @@
     ],
     "external_dependencies": {
         "python": [
-            "zeep",
+            "zeep==3.4.0",
```
sbidoul: @mathisjacoby we should never pin dependencies here.

Why do you need to pin them? Is it for the vcr cassettes to work, or is it necessary to run de module?

mathisjacoby: That's something I did when I was trying to get my tests working with the vcr cassette. 
I changed it and it also works with a newer zeep version. 
Thanks for the review !

xavier-bouquiaux (approved):
LGTM
Thanks for the migration

sbidoul (approved):

PR #213, opened 2023-12-26
[16.0][ADD] sale_margin_tax

len-foss:
This module takes a simple approach to the Belgian margin tax.

https://finances.belgium.be/fr/entreprises/tva/assujettissement-tva/regime-imposition-marge

With this, the tax is due on the margin, that is to say the benefit made between the purchase price and the selling price of used goods (not taking into account any cost incurred in between, including refurbishing or repairing costs).

Note that the seller has to be able to track the purchasing price in their inventory.
There are different ways this can be done - by lot, serial number, retrieved from the purchase order, etc.
By default, this module assumes that the purchase price is the same as the cost encoded on the product.
This advanced behaviour can be implemented on overriding modules.

When such a tax is used, the actual tax amount has to be computed at invoicing time.
Moreover, the actual tax amount should be hidden from the report, and an obligatory note has to be present on the invoice.


len-foss:
Hello @rousseldenis!
Since you worked with margin tax :-)

In 16.0 it requires more work to hide the tax information in the report -- I'm not sure I've taken the best approach, and it is annoying as it also hides the margin tax total in the backend (maybe there needs to be a context key to know if _prepare_tax_totals is used in the context of the customer report or internally).

PR #216, opened 2024-03-01
[12.0][FIX] control vat format on vat client listing

victor-champonnois:
Belgian VAT number now can start with a 1 rather than a 0 (see https://www.easytax.co/fr/tax-mag/info/belgique-changement-de-format-pour-les-numeros-de-tva-belges/)

Also autoremove the dots in the VAT numbers (they are not checked by the VIES control, but produce errors in the VAT client listing report)

forward ported in v14 : https://github.com/OCA/l10n-belgium/pull/217

### DIFF
--- a/l10n_be_vat_reports/wizard/partner_vat_list.py
+++ b/l10n_be_vat_reports/wizard/partner_vat_list.py
@@ -41,7 +41,8 @@ class PartnerVATList(models.TransientModel):
                 _("No VAT number associated with your company.")
             )
 
-        company_vat = company_vat.replace(" ", "").upper()
+        be_id = self.env.ref("base.be").id
+        company_vat = self.env["res.partner"].fix_eu_vat_number(be_id, company_vat)
         for listing in self:
             seq_declarantnum = self.env["ir.sequence"].next_by_code(
                 "declarantnum"
@@ -67,6 +68,7 @@ class PartnerVATList(models.TransientModel):
         date_start = date(int(self.year), 1, 1)
         date_stop = date(int(self.year), 12, 31)
 
+        be_id = self.env.ref("base.be").id
         partners = self.env["partner.vat.list.client"].browse([])
         be_partners = self.env["res.partner"].search([("vat", "ilike", "BE%")])
         if not be_partners:
@@ -140,7 +142,9 @@ class PartnerVATList(models.TransientModel):
         self.env.cr.execute(query, args)
         seq = 0
         for record in self.env.cr.dictfetchall():
-            record["vat"] = record["vat"].replace(" ", "").upper()
+            record["vat"] = self.env["res.partner"].fix_eu_vat_number(
+                be_id, record["vat"]
+                )
             if record["turnover"] >= self.limit_amount:
                 seq += 1
                 record["seq"] = seq
--- a/l10n_be_vat_reports/wizard/partner_vat_list_client.py
+++ b/l10n_be_vat_reports/wizard/partner_vat_list_client.py
@@ -21,17 +21,17 @@ class VATListingClients(models.TransientModel):
     @api.constrains("vat")
     def _check_vat_number(self):
         """
-        Belgium VAT numbers must respect this pattern: 0[1-9]{1}[0-9]{8}
+        Belgium VAT numbers must respect this pattern: [0-1][0-9]{9}
         todo current code assumes vat numbers start with a two-letter
           country code
         """
-        be_vat_pattern = re.compile(r"^BE0[1-9]{1}[0-9]{8}$")
+        be_vat_pattern = re.compile(r"^BE[0-1][0-9]{9}$")
         for client in self:
             if not be_vat_pattern.match(client.vat):
                 raise ValidationError(
                     _(
                         "Belgian Intervat platform only accepts VAT numbers "
-                        "matching this pattern: 0[1-9]{1}[0-9]{8} (number "
+                        "matching this pattern: [0-1][0-9]{9} (number "
                         "part). Check vat number %s for client %s"
                     )
                     % (client.vat, client.name)


l10n_be_vat_reports/wizard/partner_vat_list.py
```diff
@@ -41,7 +41,7 @@ def _compute_declarant_reference(self):
                 _("No VAT number associated with your company.")
             )
 
-        company_vat = company_vat.replace(" ", "").upper()
+        company_vat = company_vat.replace(" ", "").replace(".", "").upper()
```
huguesdk: since this is done in 2 places, it should go into a function.

also, to avoid multiple replace operations, `re.sub()` can be used:

```python
company_vat = re.sub("[ .]", "", company_vat).upper()
```

or maybe use [fix_eu_vat_number()](https://github.com/OCA/OCB/blob/16.0/addons/base_vat/models/res_partner.py#L148) from the `base_vat` module directly?

l10n_be_vat_reports/wizard/partner_vat_list_client.py
```diff
@@ -21,17 +21,17 @@ class VATListingClients(models.TransientModel):
     @api.constrains("vat")
     def _check_vat_number(self):
         """
-        Belgium VAT numbers must respect this pattern: 0[1-9]{1}[0-9]{8}
+        Belgium VAT numbers must respect this pattern: [0-1][0-9]{9}
         todo current code assumes vat numbers start with a two-letter
           country code
         """
-        be_vat_pattern = re.compile(r"^BE0[1-9]{1}[0-9]{8}$")
+        be_vat_pattern = re.compile(r"^BE[0-1][0-9]{9}$")
```
huguesdk, line 28: the previous pattern forbid numbers starting with "00", while this does not. is this ok?

huguesdk (changes requested):

l10n_be_vat_reports/wizard/partner_vat_list_client.py
victor-champonnois: Yes, according to this [source](https://economie.fgov.be/fr/themes/entreprises/banque-carrefour-des/actualites/structure-du-numero), the format is 0xxx.xxx.xxx or 1xxx.xxx.xxx

l10n_be_vat_reports/wizard/partner_vat_list.py
victor-champonnois: yes, using fix_eu_vat_number works, and it's a better solution, thanks :)
I port this change to the 14 branch

l10n_be_vat_reports/wizard/partner_vat_list.py
```diff
@@ -140,7 +141,8 @@ def get_partners(self):
         self.env.cr.execute(query, args)
         seq = 0
         for record in self.env.cr.dictfetchall():
-            record["vat"] = record["vat"].replace(" ", "").upper()
+            be_id = self.env.ref("base.be").id
```
huguesdk: this line should be moved out of the loop.

huguesdk (approved):
just one slight change to do, squash all commits and we’ll be good to go. approving already to save time.

edit: and pre-commit should be run.

victor-champonnois:
@huguesdk done !

robinkeunen (approved):
LGTM thanks !

huguesdk:
@OCA/local-belgium-maintainers could you please merge this?

sbidoul:
Hm weird error. I'll need to look into this.

huguesdk:
@sbidoul i guess the ci on 12.0 is deprecated? can this be merged manually, then (if yes, should we increment de version before)? anyway, could you remove the stale label and add no stale?

sbidoul:
@huguesdk I had lost track of this. v12 is still supported in OCA. Could you rebase?

huguesdk:
@sbidoul the branch is already based on the latest commit of 12.0.

sbidoul:
Ah, I see. Same as https://github.com/OCA/l10n-spain/pull/3733#issuecomment-2400440210. Could you try the procedure explained there?

huguesdk:
@sbidoul i get exactly the same error:
```
(...)
xlwt==1.3.0
yarl==1.7.2
zeep==4.1.0
zope.event==4.6
zope.interface==5.5.2
+ echo addons_path=/opt/odoo/addons,.
+ cat /etc/odoo.cfg
[options]
addons_path=/opt/odoo/addons,.
++ oca_list_external_dependencies deb
could not obtain odoo.addons.__path__ using python
not found in addons path: base,base_vat_sanitized,mis_builder,partner_identification,report_xml,web_notify
Aborted.
+ deps=
```

before running `oca_install_addons`, `/etc/odoo.cfg` contains only:
```ini
[options]
```
after running it, it contains:
```ini
[options]
addons_path=/opt/odoo/addons,.
```
it doesn’t find the `base` module because it is in `/opt/odoo/odoo/addons`, which is not in the `addons_path`. the other modules that are not found are oca modules (which i guess should be automatically installed from pypi). `report_xml` and `web_notify` are from repositories that are not in `oca_dependencies.txt`, but adding them doesn’t have any effect.

sbidoul:
Sorry, I meant adding `lxml<5.2` in test-requirements.txt, like they did in l10n-spain in https://github.com/OCA/l10n-spain/pull/3762

sbidoul:
You can see the error by running `docker run -it --rm -v .:/_build --workdir /_build ghcr.io/oca/oca-ci/py3.6-odoo12.0 bash`, then `oca_install_addons`, then `odoo`.

huguesdk:
@sbidoul indeed, that did it, thanks! i created a separate branch for this (where i also addedd the missing repositories in `oca_dependencies.txt`: #233.

huguesdk:
yes, finally! many thanks, @sbidoul!

PR #217, opened 2024-03-01
[14.0][FIX] control vat format on vat client listing 

victor-champonnois:
forward port of : https://github.com/OCA/l10n-belgium/pull/216

- Belgian VAT number now can start with a 1 rather than a 0 (see https://www.easytax.co/fr/tax-mag/info/belgique-changement-de-format-pour-les-numeros-de-tva-belges/). 
- use odoo's vat autocorrect for the VAT numbers (currently dots are they are not checked by the VIES control, but produce errors in the VAT client listing report)


robinkeunen (approved):
reviewed #216 , merging this as well.

PR #210, opened 2023-11-12
[MIG] l10n_be_intrastat_product: Migration to 16.0

ghost:
Waiting for this PR to be merged: https://github.com/OCA/intrastat-extrastat/pull/246

PR #220, opened 2024-04-29
[MIG] l10n_be_intrastat_product: Migration to 16.0

luc-demeyer:
superceeds https://github.com/OCA/l10n-belgium/pull/210


luc-demeyer:
@sbidoul 
it looks like there is a problems with the runners.
Any idea why the jobs remain queued ?

sbidoul:
@luc-demeyer yes:https://github.com/OCA/oca-addons-repo-template/issues/250

PR #186, opened 2023-04-12
[16.0][ADD] New module l10n_be_bpost_address_validation

kouffsamuel:
This PR will allow to validate the partner's address with the API of bpost.

PR #227, opened 2024-10-16
[FIX] l10n_be_bpost_address_autocomplete: remove headers from api call

IT-Ideas:
Since the required headers are no more supported, this commit removes them from the api call in order to successfully fetch the results.

PR #228, opened 2024-11-07
[16.0][ADD] New module l10n_be_bpost_address_validation

lmignon:
supersedes #186


remytms:
@IT-Ideas @kouffsamuel @rousseldenis As you’ve expressing interest in the PR #186 , can you review this PR that superseed it ?

remytms:
@lmignon there is a comment from @IT-Ideas in the superseeded PR.

> @lmignon not sure if it uses the same api as the autocompletion module but if it is the case we should integrate the fix of #227

I re-post it here so that we can follow it.

remytms:
Works in production for Acsone, but as @lmignon said to me, it randomly stop working, and then works again.

remytms:
Ok, bot, I saw the link to the failed commit. ^^

remytms:
Closed in favor of #256

PR #230, opened 2024-11-28
[FIX] add test dependencies to manifest

xavier-bouquiaux:
It makes more sense to add the dependencies in the manifest <br/>
This module is installable via odoo app store and for now it does not work <br/>
Odoo run the tests when installing the module and the test dependencies are not installed <br/>
For a tech profile it is not a big issue but for non tech user it can be <br/>
Generic discussion to have : why not always include test dependencies in the manifest ?

@sbidoul

### DIFF
--- a/companyweb_base/__manifest__.py
+++ b/companyweb_base/__manifest__.py
@@ -24,6 +24,8 @@
     "external_dependencies": {
         "python": [
             "zeep",
+            "freezegun",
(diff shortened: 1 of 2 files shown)


sbidoul:
Yes, this confuses users installing the module on `odoo.sh` as odoo.sh runs all tests by default.
So by installing test dependencies automatically this will reduce support calls.

PR #229, opened 2024-11-20
[18.0][MIG] companyweb_base: Migration to 18.0

BertVGroenendael:

PR #234, opened 2025-01-15
[MIG] companyweb_base to 18

xavier-bouquiaux:
Based on #229 

Client is in a rush so i based this branch on Dynapps'one and added the small correction

All migrations credit is for Dynapps

@sbidoul


sbidoul:
Compared to #229. History *looks* good.

/ocabot merge nobump

PR #238, opened 2025-03-06
[IMP] company_web : credentials are now editable

xavier-bouquiaux:
Since now the only way to change the credentials was via an export/import of the res_users

In this PR I add the credentials in the user form view


sbidoul (approved):
LGTM, tested on runboat, the screen looks good.

I notice the users cannot change the credentials by themselves on their profile. I think this is intended?

xavier-bouquiaux:
@sbidoul 
did not think about that part
I'll add the tab in the user profile

sbidoul (approved):
Tested again. Credentials on profile work.

PR #243, opened 2025-03-12
[16.0][FIX] fix tests following Odoo change in loading test CoA

sbidoul:
- l10n_be_vat_reports: make sure to use a company with the l10n_be taxes loaded. Otherwise the `_get_tax` calls in the tests fail because the belgian taxes are not created for the test company.
- fix a couple of bugs in the l10n_be_intrastat_product tests
- pre-load l10n_generic_coa to work around odoo/odoo@d0342c8

### DIFF
--- a/l10n_be_intrastat_product/tests/test_intrastat_be.py
+++ b/l10n_be_intrastat_product/tests/test_intrastat_be.py
@@ -15,6 +15,7 @@ class TestIntrastatBe(TransactionCase):
 
         cls.company = cls.env.company
         cls.env.company.country_id = cls.env.ref("base.be")
+        cls.env.company.account_fiscal_country_id = cls.env.ref("base.be")
         cls.env.company.vat = "BE0820512013"
         cls.company.intrastat_region_id = cls.env.ref(
             "l10n_be_intrastat_product.intrastat_region_2"
@@ -113,7 +114,6 @@ class TestIntrastatBe(TransactionCase):
         )
 
     def test_be_sale_b2b(self):
-
         inv_out = self.inv_obj.with_context(default_move_type="out_invoice").create(
             {
                 "partner_id": self.partner_b2b_1.id,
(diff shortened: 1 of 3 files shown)


pedrobaeza:
We already passed by all these tests, and finishing implementing what we proposed in the mailing list. Can you try?

sbidoul:
I seems there were a few real bugs in the test suite. But otherwise pre-loading l10n_generic_coa seems like an easy fix for the test suite.

l10n_be_intrastat_product/tests/test_intrastat_be.py
```diff
@@ -15,6 +15,7 @@ def setUpClass(cls):
 
         cls.company = cls.env.company
         cls.env.company.country_id = cls.env.ref("base.be")
+        cls.env.company.account_fiscal_country_id = cls.env.ref("base.be")
```
xavier-bouquiaux, line 18: What's do you declare this variable ? <br/>

xavier-bouquiaux (commented):
One small the question

For the rest : LGTM

l10n_be_intrastat_product/tests/test_intrastat_be.py
sbidoul: `account_fiscal_country_id` is a compute readonly=False that defaults to country_id.

TBH, I'm not entirely sure why it started to fail without this. It failed on a consistency error in `_validate_taxes_country` on the move reversal here:

https://github.com/OCA/l10n-belgium/blob/f6aa3907845b59a4b8dff6e3efd38d106f65c5ac/l10n_be_intrastat_product/tests/test_intrastat_be.py#L154-L166

I suspect it passed by accident before.

sbidoul: Ideally these `l10n_be_intrastat_product` tests should use a new company with the BE CoA instead of hacking the country of the main company and changing the country on all the taxes as it does today.

sbidoul:
Fast-tracking this to unblock other PRs. If there are blocking comments, I'm happy to handle them afterwards.

/ocabot merge patch

PR #245, opened 2025-03-27
[18.0][MIG] l10n_be_partner_kbo_bce: Migration to 18.0

jdidderen:


jdidderen:
@xavier-bouquiaux It was my intention to look at the error but you have done it before me so thanks for the help :)

l10n_be_partner_kbo_bce/views/res_partner.xml
```diff
@@ -6,7 +6,6 @@
         <field name="arch" type="xml">
             <div name="vat_vies_container" position="after">
                 <field name="kbo_bce_number" invisible="country_code != 'BE'" />
-                <field name="country_code" invisible="1" />
```
remytms: Should’nt you check the `invisible` condition of the field `kbo_bce_number` as the `country_code` field is removed from the view and from the model ?

remytms (changes requested):

l10n_be_partner_kbo_bce/views/res_partner.xml
jdidderen: Since 18.0, it's not needed anymore. (https://github.com/odoo/odoo/pull/137031)
image

remytms: Hooo, thanks. :)

remytms (approved):
LGTM

luc-demeyer (changes requested):
Can't we drop this module, ?
cf. https://github.com/odoo/odoo/blob/4d11ba11d36a31cec6f63042c458a26582fcdd0a/addons/l10n_be/models/res_partner.py#L13

l10n_be_partner_kbo_bce/models/res_partner.py
```diff
@@ -0,0 +1,149 @@
...
+        inverse=lambda s: s._inverse_identification(
+            "kbo_bce_number", "l10n_be_kbo_bce"
+        ),
+        search=lambda s, *a: s._search_identification("l10n_be_kbo_bce", *a),
+    )
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            if vals.get("is_company"):
```
glitchov: The field is_company is not yet available at create time and i think we should use company_type.

glitchov (changes requested):

l10n_be_partner_kbo_bce/models/res_partner.py
luc-demeyer: @glitchov 
Please have a look at my comment.
Imho this module doesn't add value any more since the functionality is now part of standard Odoo OC.

glitchov: The field is there but the sync is only one way...

jdidderen: @glitchov I have adapted the create to check both fields.

jdidderen:
> Can't we drop this module, ? cf. https://github.com/odoo/odoo/blob/4d11ba11d36a31cec6f63042c458a26582fcdd0a/addons/l10n_be/models/res_partner.py#L13

@luc-demeyer I think it's up to @OCA/local-belgium-maintainers to decide what to do with the module. The standard approach and the OCA approach are not the same so it would be legitimate to maintain the module.

sbidoul:
Hello,

To me it does not make sense to have two fields that have the same meaning (`kbo_bce_number`, and the upstream `company_registry`).

So I think we should drop this module or, if there are missing features in Odoo 18, refactor this module to use the official field.

solomonprabu:
@jdidderen I would like to work on this PR, what exactly is blocking this PR? can you list the issues?

jdidderen:
> @jdidderen I would like to work on this PR, what exactly is blocking this PR? can you list the issues?

Nothing blocking this PR. Odoo now covers the case by filling the company_registry field automatically based on the VAT number. And also two maintainers agreed to say this module should be dropped.

PR #256, opened 2025-09-15
[16.0][ADD] l10n_be_bpost_address_validation

remytms:
This superseeds #228


remytms:
It seams that the issue for the CI comes from the `l10n_be_intrastat_product` module.

PR #260, opened 2025-10-13
[16.0][FIX] l10n_be_mis_reports_xml: fix vat declaration data extraction

huguesdk:
make data extraction for xml export work regardless of changes in row and cell naming in mis_builder 16.0.5.3.0. (see changes [here](https://github.com/OCA/mis-builder/pull/676/files#diff-47080a2f9b9511ea77d42be1af586127872382ce2cd198c69e11f195612fe5c0R47).)

note: tests are failing because `l10n_be_intrastat_product` did not yet adapt to changes from OCA/intrastat-extrastat#306 (fixed in #261).

### DIFF
--- a/l10n_be_mis_reports_xml/wizards/be_vat_declaration_wizard.py
+++ b/l10n_be_mis_reports_xml/wizards/be_vat_declaration_wizard.py
@@ -77,17 +77,18 @@ class BeVATDeclarationWizard(models.TransientModel):
 
     def prepare_xml_data(self):
         self.ensure_one()
-        kpi_matrix_dict = self.mr_instance_id.compute()
-        data = {
-            row["row_id"]: row["cells"][0]["val"] for row in kpi_matrix_dict["body"]
-        }
-
-        data = {
+        report = self.mr_instance_id.report_id
+        data = report.evaluate(
+            report._prepare_aep(self.mr_instance_id.company_id),
+            self.mr_instance_id.date_from,
+            self.mr_instance_id.date_to,
+        )
+        xml_data = {
             k: "{:.2f}".format(round(v, 2))
             for k, v in data.items()
             if k.startswith("grid") and v
         }
-        return data
+        return xml_data
 
     def compute_declarant_reference(self):
         return self.env["ir.sequence"].next_by_code("be.vat.declaration.declarant")


l10n_be_mis_reports_xml/wizards/be_vat_declaration_wizard.py
```diff
@@ -77,17 +79,22 @@ def generate_xml(self):
...
+        for row in kpi_matrix_dict["body"]:
+            cell = row["cells"][0]
+            value = cell["val"]
+            if not value:
+                continue
+            kpi_id = KpiMatrix._unpack_cell_id(cell["cell_id"])[0]
+            grid_kpi = grid_kpis_by_id.get(kpi_id)
+            if grid_kpi is not None:
+                xml_data[grid_kpi.name] = "{:.2f}".format(round(value, 2))
+        return xml_data
```
sbidoul, line 91: For this it's likely better to use `period_id.evaluate()`. That method should be easier to use to obtain kpi values, while `compute()` is more dedicated to rendering.

huguesdk: thanks for the review, @sbidoul! i didn’t find an `.evaluate()` method on `mis.report.instance.period`, but i changed the code to use `mis.report.evaluate()`. does it look correct to you? it gives the same results as before.

sbidoul: Yes that's the one. This should be more robust like this.

Now you can probably even simplify more by getting rid of `mr_instance_id`, getting the `report` from it's well-known xml id, and the dates from the wizard.

sbidoul (approved):
LGTM (code review). Possible improvement as a suggestion only.

l10n_be_mis_reports_xml/wizards/be_vat_declaration_wizard.py
huguesdk: i’m not very familiar with `mis_builder`. from what i see, in this use-case, the `mis.report.instance` record is first created to define the report: it uses a template and defines a start and end date. it also allows to preview, print and export the data.

if i understand correctly your suggestion, it would be to change the way the user is using this module and only use a wizard to generate the xml data (instead of first creating the `mis.report.instance`. is that correct? if it is, i prefer to leave it like it is to not change the user experience.

huguesdk: also: is the `._prepare_aep()` call correct? shouldn’t there be more arguments?

sbidoul: Yes this looks correct. I have not tested the PR, though.

jbaudoux (approved):
Tested

PR #266, opened 2025-12-20
[19.0][MIG] migrate and improve companyweb_base

tobiaszehntner:


companyweb_base/models/res_company.py
```diff
@@ -0,0 +1,29 @@
+# Copyright 2025 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    cweb_login = fields.Char("Companyweb Login")
+    cweb_password = fields.Char("Companyweb Password")
```
sbidoul: Add security cweb_download security group on these two fields.

companyweb_base/models/res_partner.py
```diff
@@ -0,0 +1,821 @@
...
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    # Additional partner fields
+    rsin_number = fields.Char(
+        "RSIN",
+        help="RSIN: 'Rechtspersonen en Samenwerkingsverbanden "
+        "Informatienummer' is used to exchange data with other government "
+        "organisations, such as the Netherlands Tax Administration. (Netherlands only)",
+    )
```
sbidoul: Does this field (or similar) exist in l10n_nl.

sbidoul: No. But let's remove this and keep `cweb_rsin_number` only.

companyweb_base/models/res_partner.py
```diff
@@ -0,0 +1,821 @@
...
+                },
+            }
+        return {
+            "type": "ir.actions.client",
+            "tag": "display_notification",
+            "params": params,
+        }
+
+    @api.model
+    def cron_companyweb_followup(self):
```
sbidoul:
```suggestion
    def _cron_companyweb_followup(self):
```

companyweb_base/models/res_company.py
sbidoul: We need a neutralize.sql for these two fields

companyweb_base/models/res_partner.py
```diff
@@ -0,0 +1,821 @@
...
+            else:
+                partner.cweb_error = False
+                partner_list.append(
+                    {
+                        SERVICE_KEYS["country_code"]: country_code,
+                        SERVICE_KEYS["identifier"]: vat or registry,
+                        SERVICE_KEYS["identifier_type"]: SERVICE_KEYS["vat"]
+                        if vat
+                        else SERVICE_KEYS["registry"],
+                        SERVICE_KEYS["reference"]: partner.id,
```
sbidoul: Use a uuid4 as followup reference, stored on the partner in a new `cweb_followup_reference` field.

companyweb_base/models/res_partner.py
```diff
@@ -0,0 +1,827 @@
...
+                len(partner_datas),
+                remaining,
+            )
+
+            updated_partners = []
+            updated_existing_partners = []
+            for partner_data in partner_datas:
+                cweb_sync_reference = (
+                    partner_data.get(SERVICE_KEYS["push_reference"], {}) or {}
+                ).get(SERVICE_KEYS["value"], "0")
```
sbidoul: A little doubt here: what does the `0` value means as default? Can it happen in practice? What does it mean to `search([("cweb_sync_reference", "=", "0")])` in that case?

Also do we need an index on cweb_sync_reference?

tobiaszehntner: sorry that was a remnant from the `id` identifier. We can avoid searching if we don't have a reference.

```
                cweb_sync_reference = (
                    partner_data.get(SERVICE_KEYS["push_reference"], {}) or {}
                ).get(SERVICE_KEYS["value"], False)
                values = Partner._cweb_parse(partner_data)
                if cweb_sync_reference and (partner := Partner.search(
                    [("cweb_sync_reference", "=", cweb_sync_reference)], limit=1
                )):
```

regarding index, would that help?
```
cweb_sync_reference = fields.Char(index='btree_not_null')
```

sbidoul: Sound good.

tobiaszehntner:
@sbidoul Companyweb has tested and approved.

sbidoul (approved):

PR #268, opened 2026-03-12
[19.0][IMP] companyweb_base: Update images/description

tobiaszehntner:


tobiaszehntner:
@xavier-bouquiaux @sbidoul I added an image called `banner.png` as I've seen in other modules. But I didn't find any documentation on how the Odoo app store chooses the main picture (or if it's configurable somewhere..?). In lack of knowing more, should we push and try?

sbidoul:
Ah, I *think* it uses the `images` key from the manifest.

tobiaszehntner:
> Ah, I _think_ it uses the `images` key from the manifest.

bien vu!

sbidoul:
I added application=True in the manifest.

/ocabot merge patch

PR #273, opened 2026-04-09
[17.0][PORT] 271 from 16.0

sbidoul:
Port #271 from 16.0 to 17.0.


sbidoul:
Retrying after #274 

/ocabot merge patch

PR #274, opened 2026-04-09
[17.0] companyweb_payment_info: fix test

sbidoul:


sbidoul:
Remove an implicit dependency on account_peppol in the test suite.

sbidoul:
Fast tracking this trivial test fix without functional change.

/ocabot merge patch

PR #275, opened 2026-05-14
[18.0][MIG] Takeover l10n_be_partner_kbo_bce: Migration to 18.0

solomonprabu:
Supersedes https://github.com/OCA/l10n-belgium/pull/245.

PR Changes
- Refactoring field `kbo_bce_number` to upstream `company_registry`.

@jdidderen  Thanks for putting the initial effort into this migration.


sbidoul:
The same question remains as in #245. Why adding a new field when `company_registry` is now standard.

solomonprabu:
> The same question remains as in #245. Why adding a new field when `company_registry` is now standard.

I thought refactoring the module to use `company_registry` instead of a new field(`kbo_bce_number`) was the issue. Now I understand that the module itself is redundant and the upstream field `company_registry`  covers the functionality.


> [!NOTE]
>  Module: `l10n_be_partner_kbo_bce`
> Status: **Obsolete**

PR #281, opened 2026-07-14
[FIX] companyweb_base: Prefer VAT country over address country

tobiaszehntner:


xavier-bouquiaux (approved):
Problem fixed, thanks for that

xavier-bouquiaux (approved):

sbidoul:
Is there a need to migrate / recompute existing data?

PR #282, opened 2026-07-14
[FIX] Guard against balance_data_dict_parent being None

Sibert-Aerts:
The diff makes the change seem more involved than it is, this fix just moves and indents two small blocks to add this check before them:
```py
if balance_data[NESTED_KEYS["balance_data_dict_parent"]]:
```
As the two blocks assume that this balance_data value can't be `None`, which it actually can be as we have observed with one of our customers, resulting in an ugly `TypeError: 'NoneType' object is not subscriptable` exception.

### DIFF
--- a/companyweb_base/cweb_utils.py
+++ b/companyweb_base/cweb_utils.py
@@ -110,25 +110,26 @@ def get_balance_values(data):
         }
     )
     if balance_enable and balance_data:
-        balance_details = {
-            d[NESTED_KEYS["balance_data_key"]]: d[NESTED_KEYS["balance_data_value"]]
-            for d in balance_data[NESTED_KEYS["balance_data_dict_parent"]][
-                NESTED_KEYS["balance_data_dict_child"]
-            ]
-        }
-
         values.update(
             {
                 balance_field: balance_data[NESTED_KEYS[balance_field]]
                 for balance_field in BALANCE_FIELDS
             }
         )
-        values.update(
-            {
-                balance_field: balance_details[NESTED_KEYS[balance_field]]
-                for balance_field in BALANCE_NESTED_FIELDS
+
+        if balance_data[NESTED_KEYS["balance_data_dict_parent"]]:
+            balance_details = {
+                d[NESTED_KEYS["balance_data_key"]]: d[NESTED_KEYS["balance_data_value"]]
+                for d in balance_data[NESTED_KEYS["balance_data_dict_parent"]][
+                    NESTED_KEYS["balance_data_dict_child"]
+                ]
             }
-        )
+            values.update(
+                {
+                    balance_field: balance_details[NESTED_KEYS[balance_field]]
+                    for balance_field in BALANCE_NESTED_FIELDS
+                }
+            )
     else:
         # Set balance fields to False if not available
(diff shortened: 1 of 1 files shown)


sbidoul:
@tobiaszehntner

tobiaszehntner (approved):
Thanks! LGTM

companyweb_base/cweb_utils.py
```diff
@@ -110,25 +110,26 @@ def get_balance_values(data):
...
                 balance_field: balance_data[NESTED_KEYS[balance_field]]
                 for balance_field in BALANCE_FIELDS
             }
         )
-        values.update(
-            {
-                balance_field: balance_details[NESTED_KEYS[balance_field]]
-                for balance_field in BALANCE_NESTED_FIELDS
+
+        if balance_data[NESTED_KEYS["balance_data_dict_parent"]]:
```
xavier-bouquiaux, line 120: what if the data is no more available ? 
Shouldn't you have a else statement to clean the field ?

xavier-bouquiaux (commented):
Thanks a lot for your contribution

Just one small question, maybe the same case is present elsewhere in the code but as you worked on this topic i would appreciate your feedback

companyweb_base/cweb_utils.py
tobiaszehntner: True. Maybe like this
```
        values.update(
            {
                balance_field: balance_data[NESTED_KEYS[balance_field]]
                for balance_field in BALANCE_FIELDS
            }
        )
        balance_data_parent = balance_data[NESTED_KEYS["balance_data_dict_parent"]]
        if balance_data_parent:
            balance_details = {
                d[NESTED_KEYS["balance_data_key"]]: d[NESTED_KEYS["balance_data_value"]]
                for d in balance_data_parent[NESTED_KEYS["balance_data_dict_child"]]
            }
            values.update(
                {
                    balance_field: balance_details[NESTED_KEYS[balance_field]]
                    for balance_field in BALANCE_NESTED_FIELDS
                }
            )
        else:
            values.update(
                {balance_field: False for balance_field in BALANCE_NESTED_FIELDS}
            )
```

Sibert-Aerts:
> Just one small question, maybe the same case is present elsewhere in the code but as you worked on this topic i would appreciate your feedback

Hi Xavier, I have absolutely no knowledge of the Companyweb API or even exactly what the integration does, I just chased down an exception one of our clients encountered and found this apparent problem in this specific bit of code by debugging and comparing the code to the Companyweb response data. Fixing this one bit was luckily all that was needed to resolve the issue for our client.

I assume you have access to the Companyweb API docs (which aren't public as far as I could see...) so maybe you could indeed check if this field was stated to be NULL-able and thus the code was a mistake, and cross-reference if similar other fields were also assumed not to be NULL-able elsewhere in code...

I'll also leave any further changes to the maintainers since again I'm fairly clueless about Companyweb, I hope that's OK!

xavier-bouquiaux (approved):
As discussed, the cleaning of the field that would not be present anymore will be analysed outside of this issue
the main goal was to fix a bug and it's done
Thanks you

